PR #6, opened 2024-04-27
Notation section review

davxy:


text/notation.tex
```diff
@@ -85,17 +84,17 @@ \subsection{Dictionaries}\label{sec:dictionaries}
...
 
-To denote the active domain (\ie set of keys) of a dictionary $\mathbf{d} \in \dict{K}{V}$, we use $\keys{\mathbf{d}} \subset K$ and for the range (\ie set of values), $\mathcal{V}(\mathbf{d}) \subset V$. Formally:
+To denote the active domain (\ie set of keys) of a dictionary $\mathbf{d} \in \dict{K}{V}$, we use $\keys{\mathbf{d}} \subseteq K$ and for the range (\ie set of values), $\mathcal{V}(\mathbf{d}) \subseteq V$. Formally:
 \begin{align}
   \keys{\mathbf{d} \in \mathbb{D}} &\equiv \{\ k \mid \exists v : (k \mapsto v) \in \mathbf{d}\ \} \\
   \mathcal{V}(\mathbf{d} \in \mathbb{D}) &\equiv \{\ v \mid \exists k : (k \mapsto v) \in \mathbf{d}\ \}
 \end{align}
 
-Note that since the domain of $\mathcal{V}$ is a set, should different keys with equal values appear in the dictionary, the set will only contain one such value.
+Note that since the output of $\mathcal{V}$ is a set, should different keys with equal values appear in the dictionary, the set will only contain one such value.
```
davxy: I'm not 100% sure I understood the original semantic here. But if as input we have a dictionary like: 

d = {(k1 → v1), (k2 → v2), (k3 → v2) }

Then V(d) = { v1, v2 }

That is, in the output we squash equal values because it is a set

text/notation.tex
```diff
@@ -28,8 +28,7 @@ \subsection{Functions and Operators}\label{sec:functions}
 
 The substitute-if-nothing function $\mathcal{U}$ is equivalent to the first argument which is not $\none$, or $\none$ if no such argument exists:
 \begin{align}\label{eq:substituteifnothing}
-%  \mathcal{U}(a_0, \dots a_n ) \equiv a_x : (a_x \ne \none \vee x = n), \bigwedge_{i=0}^{x-1} a_i = \none
-  \mathcal{U}(a_0, \dots) \equiv a_x : (a_x \ne \none \vee x = n), \bigwedge_{i=0}^{x-1} a_i = \none
+  \mathcal{U}(a_0, \dots a_n ) \equiv a_x : (a_x \ne \none \vee x = n), \bigwedge_{i=0}^{x-1} a_i = \none
```
davxy, line 31: The commented statement seems to be correct, right?

text/notation.tex
```diff
@@ -85,17 +84,17 @@ \subsection{Dictionaries}\label{sec:dictionaries}
...
   \mathcal{V}(\mathbf{d} \in \mathbb{D}) &\equiv \{\ v \mid \exists k : (k \mapsto v) \in \mathbf{d}\ \}
 \end{align}
 
-Note that since the domain of $\mathcal{V}$ is a set, should different keys with equal values appear in the dictionary, the set will only contain one such value.
+Note that since the output of $\mathcal{V}$ is a set, should different keys with equal values appear in the dictionary, the set will only contain one such value.
 
 \subsection{Tuples}\label{sec:tuples}
 
-Tuples are groups of values where each item typically belongs to a different set. They are denoted with parentheses, \eg the tuple $t$ of the integers $3$ and $5$ is denoted $t = (3, 5)$, and it exists in the set of integer pairs sometimes denoted $\N \times \N$, but denoted in the present work as $(\N, \N)$.
+Tuples are groups of values where each item may belong to a different set. They are denoted with parentheses, \eg the tuple $t$ of the integers $3$ and $5$ is denoted $t = (3, 5)$, and it exists in the set of integer pairs sometimes denoted $\N \times \N$, but denoted in the present work as $(\N, \N)$.
```
davxy, line 97: "may" (maybe) fits better. 
E.g. the given example uses two equal sets (N, N)

text/notation.tex
```diff
@@ -56,7 +55,7 @@ \subsection{Numbers}\label{sec:numbers}
 
 \subsection{Dictionaries}\label{sec:dictionaries}
 
-A \emph{dictionary} is a possibly partial mapping from some domain into some co-domain in much the same manner as a regular function. Unlike functions however, with dictionaries the total set of pairings are necessarily enumerable, and we represent them in some data structure as the set of all $(key \mapsto value)$ pairs. (In such data-defined mappings, it is common to name the values within the range a \emph{key} and the values within the domain a \emph{value}, hence the naming.)
+A \emph{dictionary} is a possibly partial mapping from some domain into some co-domain in much the same manner as a regular function. Unlike functions however, with dictionaries the total set of pairings are necessarily enumerable, and we represent them in some data structure as the set of all $(key \mapsto value)$ pairs. (In such data-defined mappings, it is common to name the values within the domain a \emph{key} and the values within the range a \emph{value}, hence the naming.)
```
gavofyork:
```suggestion
A \emph{dictionary} is a possibly partial mapping from some domain into some co-domain in much the same manner as a regular function. Unlike functions however, with dictionaries the total set of pairings are necessarily enumerable, and we represent them in some data structure as the set of all $(key \mapsto value)$ pairs. (In such data-defined mappings, it is common to name the values within the domain a \emph{key} and the values within the co-domain a \emph{value}, hence the naming.)
```

PR #7, opened 2024-04-29
Fix Merklization function definition

zdave-parity:
Missing recursion.


zdave-parity:
Also looks like:

- Incorrect bit ranges in the B function definition (fixed by this PR).
- The "slicing" notation does not match what is defined in the notation section -- in the notation section slicing is defined using an ellipsis rather than a colon? Haven't done anything about that as not sure which side should be corrected.

gavofyork:
Yeah, you're right - the `:` should be a `\dots`.

PR #21, opened 2024-06-19
Fixes from Singapore PBA Lecture

shawntabrizi:
- Noted in the lecture that N sub V should use san-serif V. Replaced other such instances assuming the same.
- Noted in the lecture that `k_e` is incorrect, and should be `k_b` referencing the bandersnatch keys
- Number of cores should use san-serif `C`
- Removes a hanging apostrophe in "Jugements > Extrinsics"
- Removes `\seq{\H_B}_{2:3}` from the definition of J, the sequence of judgements introduced in the block’s extrinsic. (As noted in class)

### DIFF
--- a/text/header.tex
+++ b/text/header.tex
@@ -41,7 +41,7 @@ We assume the state-Merklization function $\mathcal{M}_\sigma$ is capable of tra
 
 All blocks have an associated public key to identify the author of the block. We identify this as an index into the current validator set $\kappa$. We denote the Bandersnatch key of the author as $\mathbf{H}_a$ though note that this is merely an equivalence, and is not serialized as part of the header.
 \begin{equation}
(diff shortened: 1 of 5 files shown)


shawntabrizi:
It was also noted that perhaps `\N_T` should be `\N_\mathsf{T}` in "6.1. Timekeeping", but I saw no instances of this notation so I am not certain of the correctness of this change.

gavofyork:
> It was also noted that perhaps `\N_T` should be `\N_\mathsf{T}` in "6.1. Timekeeping" (and perhaps the other instances of `\N_T`), but I saw no evidence of this notation anywhere in the paper, so I am not certain of the correctness of this change.
> 
> This change is not included in this PR.

Yes - this was correct after all.

gavofyork (approved):

PR #34, opened 2024-07-04
Various fixes for Elves security

gavofyork:
- [x] Don't immediately alter kappa mid-epoch as it affects off-chain judgements.
- [x] Instead, apply the blacklist to guarantor sig verification directly.
- [x] Include the core in the WR.
- [x] Deposit the WR's signatures in with the judgement.
- [x] Require at >= 1 negative judgements to be included with a positive verdict, and place signer in punish keys.
- [x] Serialization of judgement stuff
- [x] Should always be possible to submit more guarantee signatures of a known-bad report to place them in the punish set.
  - [x] Only from lambda and kappa
- [x] Use posterior kappa for everything except judgements.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,17 +8,6 @@ https://graypaper.com/
 
 ## Remaining near-term
 
-### ELVES
-- [x] Don't immediately alter kappa mid-epoch as it affects off-chain judgements.
-- [x] Instead, apply the blacklist to guarantor sig verification directly.
-- [x] Include the core in the WR.
-- [x] Deposit the WR's signatures in with the judgement.
-- [x] Require at >= 1 negative judgements to be included with a positive verdict, and place signer in punish keys.
-- [x] Serialization of judgement stuff
-- [x] Should always be possible to submit more guarantee signatures of a known-bad report to place them in the punish set.
-  - [x] Only from lambda and kappa?
-- [x] Use posterior kappa for everything except judgements.
-
 ### Finesse
 - [ ] Make all subscript names capitalized
 - [ ] Ensure all definitions are referenced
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -199,5 +199,15 @@ https://graypaper.com/
...
+### ELVES
+- [x] Don't immediately alter kappa mid-epoch as it affects off-chain judgements.
+- [x] Instead, apply the blacklist to guarantor sig verification directly.
+- [x] Include the core in the WR.
+- [x] Deposit the WR's signatures in with the judgement.
+- [x] Require at >= 1 negative judgements to be included with a positive verdict, and place signer in punish keys.
+- [x] Serialization of judgement stuff
+- [x] Should always be possible to submit more guarantee signatures of a known-bad report to place them in the punish set.
+  - [x] Only from lambda and kappa?
+- [x] Use posterior kappa for everything except judgements.
```
AlistairStewart, line 211: I'm not sure what's going on with assurances.

In Polkadot, at least before asynchronous backing, we timed out backers pending availability at the end of a session. This was because parablock validity depended on which session we are in.

If JAM allows WPs guaranteed in the previous epoch to be assured in this one, then which set does the voting?

I'd have thought the prior set makes sense, especially if there is no block in the last slot of the previous epoch.

text/judgements.tex
```diff
@@ -12,70 +12,92 @@ \section{Judgements}\label{sec:judgements}
 
 \subsection{State}
 
-The judgements state includes three items, an allow-set ($\psi_\mathbf{a}$), a ban-set ($\psi_\mathbf{b}$) and a punish-set ($\psi_\mathbf{p}$). The allow-set contains the hashes of all work-reports which were disputed and judged to be accurate. The ban-set contains the hashes of all work-reports which were disputed and whose accuracy could not be confidently confirmed. The punish-set is a set of Ed25519 keys representing validators which were found to have guaranteed a report which was confidently found to be invalid.
+The judgements state includes three items, an allow-set ($\psi_\mathbf{a}$), a ban-set ($\psi_\mathbf{b}$), a corrupt-set ($\psi_\mathbf{c}$) and a punish-set ($\psi_\mathbf{p}$). The allow-set, ban-set and corrupt-set contains the hashes of all work-reports which were respectively judged to be correct, uncertain or incorrect. The punish-set is a set of Ed25519 keys representing validators which were found to have misjudged a work-report.
```
AlistairStewart, line 15: So the punish set now includes evryone who misjudged and not just backers. Good, because all these validators should not be guarentors.

The staking chain will want to make a distinction to slash guarantors and others by different amounts, which liekely means that we have to repost the judgement on it, rather than relying on JAM's punish set. Tis is workable and maybe there are judgements that only go to that chain.

AlistairStewart (approved):
It looks good as far as I can tell.

PR #38, opened 2024-07-11
Judgement set has four items

vikiival:
IMG_8717


vikiival:
See that @gavofyork fixed it by himself.

understandable-have-nice-day

PR #51, opened 2024-07-24
Update PVM instruction decoding equations

koute:
This PR updates the instruction decoding equations for PVM.

These match with what I have currently implemented.

~~There's one more discrepancy that I have not addressed: in `A.5.8.` the upper bits of the byte are used as a destination register (this is inverse of what I have implemented) and in `A.5.11.` the destination register is in the lower bits of the byte (which matches what I have implemented). Ultimately it doesn't matter which one we pick, but it'd be nice to have it be consistent, so we should either adjust how GP defines `A.5.8.`, or we should adjust how GP defines `A.5.11.` and I'll change my implementation to match.~~ I've swapped the arguments.

### DIFF
--- a/text/pvm.tex
+++ b/text/pvm.tex
@@ -224,7 +224,7 @@ We assume the skip length $\ell$ is well-defined:
 \subsubsection{Instructions with Arguments of One Immediate}
 \begin{equation}
 \begin{aligned}
-  \using l_X = \max(\ell, 4) \,,\quad
+  \using l_X = \min(4, \ell) \,,\quad
   \immed_X \equiv \sext_{l_X}(\de_{l_X}(\instructions_{\imath+1\dots+l_X}))
 \end{aligned}
 \end{equation}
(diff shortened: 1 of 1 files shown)


text/pvm.tex
```diff
@@ -312,13 +314,15 @@ \subsubsection{Instructions with Arguments of One Register \& One Immediate}
...
-    \using l_X &= \max(4, \ffrac{\instructions_{\imath+1}}{16}) \,,\quad&
-    \immed_X &= \sext_{l_X}(\de_{l_X}(\instructions_{\imath+2\dots+l_X})) \\
-    \using l_Y &= \max(4, (\ell - l_X - 1) \bmod 2^8) \,,\quad&
-    \immed_Y &= \sext_{l_Y}(\de_{l_Y}(\instructions_{\imath+2+l_X\dots+l_Y}))
+    \using l_A &= (\ffrac{\instructions_{\imath+1}}{16} \bmod 8) \,,\quad&
+    \using l_X &= \min(4, \max(0, l_A)) \\
+    \using l_{Y_1} &= \min(4, \max(0, \ell - l_A - 1)) \,,\quad&
+    \using l_{Y_2} &= \min(l_{Y_1}, 8 - l_A) \\
+    \immed_X &\equiv \sext_{l_X}(\de_{l_X}(\instructions_{\imath+2\dots+l_X})) \,,\quad&
+    \immed_Y &\equiv \sext_{l_{Y_1}}(\de_{l_{Y_2}}(\instructions_{\imath+2+l_A\dots+l_{Y_2}}))
```
gavofyork: Why is this so much more complex?

gavofyork (approved):

PR #67, opened 2024-08-26
Serialization adjustments

davxy:
- Fix serialization order in eq 283 (work report) to be coherent with eq 117:
    - 117: (s, x, c, a, o, r)
    - 283: WAS (a, c, o, x, s, r)
 
  The serialization order of items in the other types is consistent with the definitions of those types, so this may be a leftover

- Fix naming for work items entry in eq 284 ($x_i -> x_w$) to be coherent with eq 174

- Adjust naming in eq 277 of dispute verdicts to be coherent with naming used in eq 97/98
	I.e. Dispute extrinsic is now $E_D ≡ (v,c,f)$ and $(r,a,j) ∈ v$

- Encode verdict age as $E_4$

---

- **Not related to serialization**: Add hashing algorithm argument to MMR accumulate call

### DIFF
--- a/text/recent_history.tex
+++ b/text/recent_history.tex
@@ -16,7 +16,7 @@ We define an item $n$ comprising the new block's header hash, its accumulation-r
 \begin{equation}
   \begin{aligned}
     \using r &= \mathcal{M}_B(\orderby{s}{\se_4(s)\frown\se(h) \mid (s, h) \in \mathbf{C}}, \mathcal{H}_K) \\
-    \using \mathbf{b} &= \mathcal{A}(\text{last}(\sq{\sq{}} \concat \sq{x_\mathbf{b} \mid x \orderedin \beta}), r)\\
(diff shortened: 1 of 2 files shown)


basedafdev:
As mentioned by @vekexasia earlier in [https://github.com/w3f/jamtestvectors/pull/12], for the Assurances Extrinsic, can we also add explicit $E_2$ for the validator index?

davxy:
If variable length encoding is solely employed for the sequence length discriminant, perhaps the most straightforward approach would be to define a single rule that encodes any $\mathbb{N}_k$ as $E_k$. 

Assuming this statement remains true (GP just below eq. 272):

> Note that at present this is utilized only in encoding the length prefix of variable-length sequences.

then adopting a "catch all" rule might be easier

gavofyork:
That doesn't work in the axiomatic set theory in which GP is written since $\mathbb{N}_{2^8} \subset \mathbb{N}_{2^{16}}$, thereby making encoded integers variable length without introducing a discriminator.

To make it work we'd need to introduce a form of typing, which I'd prefer not to do as it would severely complicate the the base language of the GP.

PR #77, opened 2024-09-11
Minor corrections

gavofyork:
- Fix discrepency between Phi and Phi_1
- Bandersnatch: Use padding point for bad keys in ring
- PVM: Fix Skip function
- PVM: A.6: x' -> x 
- PVM: Typo Omega_X should be Omega_Q
- PVM host calls: Val key is 336 long
- Accumulation: Memos are 128 bytes
- Reporting/Assurance: a_v should be a_f

### DIFF
--- a/text/bandersnatch.tex
+++ b/text/bandersnatch.tex
@@ -16,3 +16,5 @@ The singly-contextualized Bandersnatch Ring\textsc{vrf} proofs $\bandersnatch{r}
   \bandersnatch{r \in \Y_R}{c \in \H}{m \in \Y} \subset \Y_{784} &\equiv \{ x \mid x \in \Y_{784}, \text{verify}(r, c, m, \text{decode}(x_{\dots32}), \text{decode}(x_{32\dots})) = \top \}  \\
   \banderout{p \in \bandersnatch{r}{c}{m}} \in \H &\equiv \text{hashed\_output}(\text{decode}(x_{\dots32}) \mid x \in \bandersnatch{r}{c}{m})
 \end{align}
+
+Note that in the case a key $\H_B$ has no corresponding Bandersnatch point, then it is disregarded from the commitment $\mathcal{O}(\seq{\H_B})$.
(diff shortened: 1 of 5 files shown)


davxy:
This may introduce additional complexity when generating the proof. The prover relies on its key index within the validators list (the one on-chain), which must align with the index of its key within the committed key sequence (aka ring). 

If we skip invalid keys when constructing the ring commitment, the index shifts, resulting in an invalid proof.

One annoying solution would require the prover to know which keys preceding its own are valid, ensuring the correct index is used.

A simpler solution would be to use a padding point. If a key is invalid when committing the ring, we substitute it with the padding point. An appropriate option might be the padding point already used to pad the ring on the right (currently, since the domain supports up to 1791 validators, we already use 768 instances of the padding point).

The padding point value is documented in the Bandersnatch specification.

text/bandersnatch.tex
```diff
@@ -16,3 +16,5 @@ \section{Bandersnatch Ring VRF}\label{sec:bandersnatch}
   \bandersnatch{r \in \Y_R}{c \in \H}{m \in \Y} \subset \Y_{784} &\equiv \{ x \mid x \in \Y_{784}, \text{verify}(r, c, m, \text{decode}(x_{\dots32}), \text{decode}(x_{32\dots})) = \top \}  \\
   \banderout{p \in \bandersnatch{r}{c}{m}} \in \H &\equiv \text{hashed\_output}(\text{decode}(x_{\dots32}) \mid x \in \bandersnatch{r}{c}{m})
 \end{align}
+
+Note that in the case a key $\H_B$ has no corresponding Bandersnatch point, then it is disregarded from the commitment $\mathcal{O}(\seq{\H_B})$.
```
davxy: This may introduce additional complexity when generating the proof. The prover relies on its key index within the validators list (the one on-chain), which must align with the index of its key within the committed key sequence (aka ring). 

If we skip invalid keys when constructing the ring commitment, the index shifts, resulting in an invalid proof.

One annoying solution would require the prover to know which keys preceding its own are valid, ensuring the correct index is used. This could also be more disadvantageous for the validators further down the list, as they would need to check more keys.

A simpler solution would be to use a padding point. If a key is invalid when committing the ring, we substitute it with the padding point. An appropriate padding point is the one we already use to pad the ring on the right (currently, since the domain supports up to 1791 validators, we already use 768 instances of this padding point).

The padding point value is in the Bandersnatch specification.

gavofyork: Ok, so we just state that when constructing the ring, any keys which have no corresponding point actually just use the padding point rather than are skipped?

davxy: Yes. It may be worth specifying that the padding point value we use is defined in \cite{hosseini2024bandersnatch}, though it might already be deduced from the text above.

PR #89, opened 2024-09-30
changed 'e' subscript to uppercase

danicuki:


danicuki:
What changes do you mean besides "k_E"? I scanned all formulas and found only these instances.

PR #111, opened 2024-10-21
Simplify definition of Q

zdave-parity:
Second argument is not necessary. Also, g is a sequence, so should really compare to [] not {}.


zdave-parity:
Just noticed this, hopefully I'm not missing anything?

gavofyork:
Yeah looks about right

gavofyork (approved):

PR #121, opened 2024-10-26
Merklisation: Changes paths so they can't collide (alternative)

gavofyork:


text/merklization.tex
```diff
@@ -40,9 +40,9 @@ \subsection{Serialization}
...
-    \forall (s \mapsto \mathbf{a}) \in \delta, (h \mapsto \mathbf{v}) \in \mathbf{a}_\mathbf{s}: &&C(s, h) &\mapsto \mathbf{v} \;, \\
-    \forall (s \mapsto \mathbf{a}) \in \delta, (h \mapsto \mathbf{p}) \in \mathbf{a}_\mathbf{p}: &&C(s, h) &\mapsto \mathbf{p} \;, \\
-    \forall (s \mapsto \mathbf{a}) \in \delta, (\tup{h, l} \mapsto \mathbf{t}) \in \mathbf{a}_\mathbf{l}: &&C(s, \se_4(l) \concat (\lnot h_{4\dots})) &\mapsto \se(\var{[\se_4(x) \mid x \orderedin \mathbf{t}]})
+    \forall (s \mapsto \mathbf{a}) \in \delta, (k \mapsto \mathbf{v}) \in \mathbf{a}_\mathbf{s}: &&C(s, \se_4(2^{32}-1) \frown k_{0\dots 28}) &\mapsto \mathbf{v} \;, \\
+    \forall (s \mapsto \mathbf{a}) \in \delta, (h \mapsto \mathbf{p}) \in \mathbf{a}_\mathbf{p}: &&C(s, \se_4(2^{32}-2) \frown h_{1\dots 29}) &\mapsto \mathbf{p} \;, \\
+    \forall (s \mapsto \mathbf{a}) \in \delta, (\tup{h, l} \mapsto \mathbf{t}) \in \mathbf{a}_\mathbf{l}: &&C(s, \se_4(l) \concat \mathcal{H}(h)) &\mapsto \se(\var{[\se_4(x) \mid x \orderedin \mathbf{t}]})
```
xlc, line 45: to confirm, this is hash of the preiamge hash? any reason to do a extra hash here? the length bytes should be enough to prevent collide as u32.max -1 is no way a valid length?

gavofyork: The user could otherwise solicit two values of `h` each with the same `l` such that the two values had the same initial 28 bytes. By hashing the `h` first, we prevent this as it's not possible to generate two preimages with the same first 28 bytes of the hash.

PR #122, opened 2024-10-26
Prerequisite is a set, ensure all dependencies are known/valid when reported

### LINKED ISSUE #103: Guarantors should not get rewarded if their SR lookup is invalid
Currently we score Guarantor performance based on whether a validator's key guaranteed a report. In fact we should only score when a Work Report is accumulated. This implies keeping the guarantor keys around alongside the Work Report, inside of $\vartheta$.

This can be done through ensuring that all dependencies are satisfied immediately at the reporting stage, rather than at the pre-accumulation stage. #122 achieves this, with the exception of checking the SR lookup mapping. In order to do this the cleanest, we must integrate the WPH->SR mapping (currently in *Accumulated*) into the Recent Blocks. Alternatively, we can keep *Accumulated* as it is but update it during the reporting process.

gavofyork:
Fixes #103 

- The WP/WR prerequisite is a set rather than an optional value.
- The size of the set of a WR's prerequisites plus the size of said WR's SR lookup dictionary must total no more than 8.
- All dependencies (defined as the set of prerequisites in union with the keys of the SR lookup dictionary) must have been recently reported at the time of reporting.
- A mapping of WPH->SR is no longer retained in Accumulated.
- A mapping of WPH->SR is maintained now in Recent Blocks ($\beta$).
- SR lookups are validated immediately upon reporting. It is up to the block producer to ensure that they are legit since a block is invalid if the SR lookups cannot be validated.

### DIFF
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -207,7 +207,8 @@ This introduces $\mathbb{O}$, the set of wrangled \emph{operand tuples}, used as
 
 Given the result of the top-level $\Delta_+$, we may define the posterior state $\chi'$, $\varphi'$ and $\iota'$ as well as the second intermediate state of the service-accounts $\delta^\ddagger$ and the \textsc{Beefy} commitment map $\beefycommitmap$:
 \begin{align}
-  \using (n, \mathbf{o}, \mathbf{t}, \beefycommitmap) &= \Delta_+(\mathsf{G}_T, \mathbf{W}^*, (\chi, \delta^\dagger, \iota, \varphi), \chi_\mathbf{g}) \\
+  \using g &= \max\left(\mathsf{G}_T, \mathsf{G}_A\cdot \mathsf{C} + \sum_{g \in \mathcal{V}(\chi_\mathbf{g})}(g)\right)\\
+  \using (n, \mathbf{o}, \mathbf{t}, \beefycommitmap) &= \Delta_+(g, \mathbf{W}^*, (\chi, \delta^\dagger, \iota, \varphi), \chi_\mathbf{g}) \\
   (\chi', \delta^\ddagger, \iota', \varphi') &\equiv \mathbf{o}
 \end{align}
 
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -39,7 +39,7 @@ We define $\delta^\dagger$ as the state after the integration of the preimages:
 
 \subsection{History and Queuing}
 
-Accumulation of a work-package/work-report is deferred in the case that it has a not-yet-fulfilled dependency and is cancelled entirely in the case of an invalid dependency. Dependencies are specified as work-package hashes and in order to know which work-packages have been accumulated already, we maintain a history of what has been accumulated. This history, $\accumulated$, is sufficiently large for an epoch worth of work-reports. For all such reports, we also retain their segment-root in order that we may validate any guarantor-proscribed segment-roots in work-reports prior to accumulation. Formally:
+Accumulation of a work-package/work-report is deferred in the case that it has a not-yet-fulfilled dependency and is cancelled entirely in the case of an invalid dependency. Dependencies are specified as work-package hashes and in order to know which work-packages have been accumulated already, we maintain a history of what has been accumulated. This history, $\accumulated$, is sufficiently large for an epoch worth of work-reports. Formally:
 \begin{align}
   \accumulated &\in \seq{\{\H\}}_\mathsf{E} \\
   \accumulatedcup &\equiv \bigcup_{x \in \accumulated}(x)
@@ -52,16 +52,16 @@ We also maintain knowledge of ready (\ie available and/or audited) but not-yet-a
 
 The newly available work-reports, $\mathbf{W}$, are partitioned into two sequences based on the condition of having zero prerequisite work-reports. Those meeting the condition, $\mathbf{W}^!$, are accumulated immediately. Those not, $\mathbf{W}^Q$, are for queued execution. Formally:
 \begin{align}
-  \mathbf{W}^! &\equiv [ w \mid w \orderedin \mathbf{W}, (w_x)_p = \none \wedge w_\wr¬srlookup = \{\} ] \\
+  \mathbf{W}^! &\equiv [ w \mid w \orderedin \mathbf{W}, (w_x)_\mathbf{p} = \none \wedge w_\wr¬srlookup = \{\} ] \\
   \mathbf{W}^Q &\equiv E(\sq{
     D(w) \mid
     w \orderedin \mathbf{W},
-    (w_x)_p \ne \none \vee w_\wr¬srlookup \ne \{\}
+    (w_x)_\mathbf{p} \ne \none \vee w_\wr¬srlookup \ne \{\}
   }, \accumulatedcup)\!\!\!\!\\
-  D(w) &\equiv (w, \{(w_x)_p\} \cup \keys{w_\wr¬srlookup})
+  D(w) &\equiv (w, \{(w_x)_\mathbf{p}\} \cup \keys{w_\wr¬srlookup})
 \end{align}
 
-We define the queue-editing function $E$, which is essentially a mutator function for items such as those of $\ready$, parameterized by a dictionary such as the items in $\accumulated$. It is used to update queues of work-reports when some of them are accumulated. Functionally, it removes all entries whose work-report's hash is in the parameter's keys, removes any dependencies which appear in said set and, crucially, removes any entries whose segment-root correspondences diverge from those in the dictionary. Formally:
+We define the queue-editing function $E$, which is essentially a mutator function for items such as those of $\ready$, parameterized by sets of now-accumulated work-package hashes (those in $\accumulated$). It is used to update queues of work-reports when some of them are accumulated. Functionally, it removes all entries whose work-report's hash is in the parameter's keys, removes any dependencies which appear in said set and, crucially, removes any entries whose segment-root correspondences diverge from those in the dictionary. Formally:
 \begin{equation}
   E\colon\left\{\begin{aligned}
       &(\seq{(\mathbb{W}, \{\H\})}, \{\H\}) \to \seq{(\mathbb{W}, \{\H\})} \\
@@ -88,7 +88,7 @@ We further define the accumulation priority queue function $Q$, which provides t
   \end{aligned}\right.
 \end{equation}
 
-Finally, we define the mapping function $\srmap$ which extracts all work-package hashes in the segment-root dictionaries of a set of work-reports:
+Finally, we define the mapping function $\srmap$ which extracts the corresponding work-package hashes from a set of work-reports:
 \begin{equation}
   \srmap\colon\left\{\begin{aligned}
     \{\mathbb{W}\} &\to \{\H\}\\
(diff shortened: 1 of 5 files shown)


text/serialization.tex
```diff
@@ -139,7 +139,7 @@ \subsection{Block Serialization}
   \nonumber &\quad \where \xtdisputes \equiv (\mathbf{v}, \mathbf{c}, \mathbf{f}) \\
   \se(\mathbf{H}) &= \se_U(\mathbf{H})\concat\se(\mathbf{H}_s) \\
   \se_U(\mathbf{H}) &= \se(\mathbf{H}_p,\mathbf{H}_r,\mathbf{H}_x)\concat\se_4(\mathbf{H}_t)\concat\se(\maybe{\mathbf{H}_e},\maybe{\mathbf{H}_w},\var{\mathbf{H}_o},\se_2(\mathbf{H}_i),\mathbf{H}_v)\\
-  \se(x \in \mathbb{X}) &\equiv \se(x_a, x_s, x_b, x_l)\concat\se_4(x_t)\concat\se(\maybe{x_p})\\
+  \se(x \in \mathbb{X}) &\equiv \se(x_a, x_s, x_b, x_l)\concat\se_4(x_t)\concat\se(\var{x_p})\\
```
zdave-parity: `p` -> `\mathbf{p}`?

text/accumulation.tex
```diff
@@ -64,14 +64,13 @@ \subsection{History and Queuing}
 We define the queue-editing function $E$, which is essentially a mutator function for items such as those of $\ready$, parameterized by a dictionary such as the items in $\accumulated$. It is used to update queues of work-reports when some of them are accumulated. Functionally, it removes all entries whose work-report's hash is in the parameter's keys, removes any dependencies which appear in said set and, crucially, removes any entries whose segment-root correspondences diverge from those in the dictionary. Formally:
```
zdave-parity: Can't put a comment on the right line (thanks GitHub), but `(w_x)_p = \none` and `(w_x)_p \ne \none` above want to be `(w_x)_{\mathbf{p}} = \{\}` and `(w_x)_{\mathbf{p}} \ne \{\}` now?

Also `\{(w_x)_p\}` should be `(w_x)_{\mathbf{p}}`.

text/accumulation.tex
```diff
@@ -41,7 +41,7 @@ \subsection{History and Queuing}
 
 Accumulation of a work-package/work-report is deferred in the case that it has a not-yet-fulfilled dependency and is cancelled entirely in the case of an invalid dependency. Dependencies are specified as work-package hashes and in order to know which work-packages have been accumulated already, we maintain a history of what has been accumulated. This history, $\accumulated$, is sufficiently large for an epoch worth of work-reports. For all such reports, we also retain their segment-root in order that we may validate any guarantor-proscribed segment-roots in work-reports prior to accumulation. Formally:
```
zdave-parity: Last bit needs updating; we don't keep the segment-roots any more.

text/accumulation.tex
zdave-parity: Paragraph needs updating to reflect accumulated now containing sets rather than dictionaries, and also that the segment-root mappings are not checked here any more.

text/accumulation.tex
```diff
@@ -89,11 +88,11 @@ \subsection{History and Queuing}
   \end{aligned}\right.
 \end{equation}
 
-Finally, we define the mapping function $\srmap$ which builds a dictionary of work-package hashes to segment-roots from a set of work-reports:
+Finally, we define the mapping function $\srmap$ which extracts all work-package hashes in the segment-root dictionaries of a set of work-reports:
```
zdave-parity: This doesn't do anything segment-root related now? It just extracts the work-package hashes from the given work-reports. Should probably rename `\srmap` for that reason, though that isn't visible in the final PDF so guess it doesn't really matter.

text/accumulation.tex
```diff
@@ -208,7 +207,8 @@ \subsection{Deferred Transfers and State Integration}
 
 Given the result of the top-level $\Delta_+$, we may define the posterior state $\chi'$, $\varphi'$ and $\iota'$ as well as the second intermediate state of the service-accounts $\delta^\ddagger$ and the \textsc{Beefy} commitment map $\beefycommitmap$:
 \begin{align}
-  \using (n, \mathbf{o}, \mathbf{t}, \beefycommitmap) &= \Delta_+(\mathsf{G}_T, \mathbf{W}^*, (\chi, \delta^\dagger, \iota, \varphi), \chi_\mathbf{g}) \\
+  \using g &= \max\left(\mathsf{G}_T, \mathsf{G}_A\cdot \mathsf{C} + \sum_{x \in \mathcal{V}(\chi_\mathbf{g})}(x)\right)\\
```
zdave-parity, line 210: The "Constants" section in the appendix says G_T "may be no smaller than" the RHS of this max?

text/recent_history.tex
```diff
@@ -2,7 +2,7 @@ \section{Recent History}\label{sec:recenthistory}
 
 We retain in state information on the most recent $\mathsf{H}$ blocks. This is used to preclude the possibility of duplicate or out of date work-reports from being submitted.
 \begin{equation}
-  \beta \in \lseq\ltuple \isa{h}{\H}\ts\isa{\mathbf{b}}{\seq{\H?}}\ts\isa{s}{\H}\ts\isa{\mathbf{p}}{\lseq\H\rseq_{:\mathsf{C}}} \rtuple\rseq_{:\mathsf{H}}
+  \beta \in \lseq\ltuple \isa{h}{\H}\ts\isa{\mathbf{b}}{\seq{\H?}}\ts\isa{s}{\H}\ts\isa{\mathbf{p}}{\dict{\H}{\H}} \rtuple\rseq_{:\mathsf{H}}
 \end{equation}
 
 For each recent block, we retain its header hash, its state root, its accumulation-result \textsc{mmr} and the hash of each work-report made into it which is no more than the total number of cores, $\mathsf{C} = 341$.
```
zdave-parity: This sentence needs fixing? "hash of each work-report" at least is not accurate as the dictionary keys are work-package hashes and the values segments-roots.

text/reporting_assurance.tex
```diff
@@ -38,17 +38,17 @@ \subsubsection{Work Report}\label{sec:workreport}
 
 % TODO: places where w \in \mathbb{W} : w_o should be replaced with w_\mathbf{o} and w_r with w_\mathbf{r}
 
-We limit the number of items in the segment-root lookup dictionary to 8, and the total serialized size of a work-report may be no greater than $\mathsf{W}_R$ bytes:
+We limit the number of (unique) items in the segment-root lookup dictionary to 8, and the total serialized size of a work-report may be no greater than $\mathsf{W}_R$ bytes:
```
zdave-parity: Should mention prerequisites in this sentence?

text/reporting_assurance.tex
```diff
@@ -38,17 +38,17 @@ \subsubsection{Work Report}\label{sec:workreport}
...
 \begin{equation}
-  \forall w \in \mathbb{W} : |w_\wr¬srlookup| \le 8\ \wedge\ |\se(w)| \leq \mathsf{W}_R
+  \forall w \in \mathbb{W} : |\keys{w_\wr¬srlookup}| + |(w_x)_\mathbf{p}| \le 8\ \wedge\ |\se(w)| \leq \mathsf{W}_R
```
zdave-parity: `\keys{}` seems redundant here?

text/reporting_assurance.tex
```diff
@@ -282,18 +282,28 @@ \subsubsection{Contextual Validity of Reports}\label{sec:contextualvalidity}
...
-  \forall p \in \mathbf{p}, \forall x \in \beta : p \not\in x_\mathbf{p}
+  \forall p \in \mathbf{p}, \forall x \in \beta : p \not\in \keys{x_\mathbf{p}}
 \end{equation}
 
-We require that the prerequisite work-package, if present, be either in the extrinsic or in our recent history:
-\begin{equation}\begin{aligned}
-  &\forall w \in \mathbf{w}, (w_x)_p \ne \none :\\
-  &\quad(w_x)_p \in \mathbf{p} \cup \{ x \mid x \in b_\mathbf{p} ,\, b \in \beta \}
-\end{aligned}\end{equation}
+We require that the prerequisite work-package, if present, and any work-packages mentioned in the segment-root lookup, be either in the extrinsic or in our recent history.
```
zdave-parity: "the prerequisite work-package" --> "prerequisite work-packages", or something like this.

text/reporting_assurance.tex
```diff
@@ -282,18 +282,28 @@ \subsubsection{Contextual Validity of Reports}\label{sec:contextualvalidity}
...
+  &\begin{aligned}
+    &\forall w \in \mathbf{w}, \forall p \in (w_x)_\mathbf{p} \cup \keys{w_\mathbf{l}} :\\
+    &\quad p \in \mathbf{p} \cup \{ x \mid x \in \keys{b_\mathbf{p}} ,\, b \in \beta \}
+  \end{aligned}
+\end{align}
+
+We require that any segment roots mentioned in the segment-root lookup be verified as correct based on our recent work-package history and the present block:
+\begin{align}
+  &\using \mathbf{p} = \{ ((g_w)_s)_h \mapsto ((g_w)_s)_e \mid g \in \xtguarantees \} \\
+  &\forall w \in \mathbf{w}: w_\mathbf{l} \subset \mathbf{p} \cup \bigcup_{b \in \beta} b_\mathbf{p}
```
zdave-parity: `subset` --> `subseteq`?

zdave-parity (commented):
Makes sense to me.

text/accumulation.tex
gavofyork: True, but it feels like a bit of a cop-out or at least foot-gun. We avoid any possible uncertainty with this change.

zdave-parity:
Hmm one thing that has just occurred to me is that with this change, it is possible for bad WPH->SR mappings to be accepted, because we are simply trusting the mappings from guaranteed reports. If you collaborate with another guarantor you can get bogus mappings in the recent blocks state without punishment by sharing bogus guarantees but refusing to send any shards out.

zdave-parity:
To elaborate a bit, I'm thinking something along the following lines could happen:

- There is a work-package A which exports some segments.
- There is a work-package B which imports some segments from A by specifying A's hash.
- There are two collaborating validators assigned to the same core who would like to tamper with the data imported by B.
- A work-report X is produced for A, gets reported on chain, and accumulated.
- Enough blocks pass such that A no longer appears in the recent blocks list.
- The collaborating validators produce a bogus work-report X' claiming to be for work-package A but with a bogus segments-root and anchor (this bogus anchor just needs to be some recent block).
- X' is shared and reported on chain. This is possible because the WP appears to be recent due to the bogus anchor, and not a duplicate as A has fallen out of the recent blocks list.
- A work-report Y is produced for B which uses the bogus segments-root for A from work-report X'. The validators producing this work-report needn't necessarily be malicious; they could be honest validators that have seen work-report X' but not X.
- Y is shared and reported on chain. Note that the bogus segments-root is accepted because it matches what was in work-report X', which is in the recent blocks list.
- The collaborating validators do not share any shards for X' and so it eventually expires.
- Y is made available, passes audit because it is valid, and gets accumulated because its dependencies are satisfied (A was previously accumulated).

gavofyork:
Yup - so to fix this we need to ensure that the WP of incoming WRs not only don't appear in the Recent Blocks list but also not the Accumulated or Ready Queue.

PR #135, opened 2024-11-08
Main alterations for 0.5

### LINKED ISSUE #81: Post-Accumulate Preimage Integration
Preimage data in the Preimages Extrinsic are currently integrated prior to accumulation. Instead they should be integrated after accumulation, but validated before accumulation. Any preimages whose integration became invalid during accumulation should be ignored but should NOT imply an invalid block.

This facilitates the parallelisation of preimage collection/integration and accumulation.

### LINKED ISSUE #82: Extrinsic Commitment should be hash of hashes
At present the Extrinsic Commitment is a hash of the extrinsic data. This should be changed to either a Merkle root of all the various extrinsic components, or, slightly simpler and more efficient, a hash of the concatenated hashes of the components.

Components at need of this are those larger items which we expect to be pre-distributed to the block:
- guarantees
- preimages

### LINKED ISSUE #94: 64-bit PVM

### LINKED ISSUE #101: Memory access (but not values) should be page-based

### LINKED ISSUE #102: Host-call for allocating and freeing memory pages

### LINKED ISSUE #107: Have the segment-count in the Availability Specifier

gavofyork:
https://hackmd.io/@hQTcJMiNR_K6HAoF_nY4qw/BkwgsG4Jye#64-bit-changes

Closes #81
Closes #82
Closes #94
Closes #101
Closes #102
Closes #107


text/overview.tex
```diff
@@ -54,17 +54,17 @@ \subsubsection{State Transition Dependency Graph}
...
-  \pi' &\prec (\xtguarantees, \xtpreimages, \xtassurances, \xttickets, \tau, \kappa', \pi, \mathbf{H})\!\!\!\!
+  \pi' &\prec (\xtguarantees, \xtpreimages, \xtassurances, \xttickets, \tau, \kappa', \pi, \mathbf{H})\!\!\!\!\!\!\!\!
 \end{align}
 
-The only synchronous entanglements are visible through the intermediate components superscripted with a dagger and defined in equations \ref{eq:betadagger}, \ref{eq:deltadagger} and \ref{eq:rhoddagger}. The latter two mark a merge and join in the dependency graph and, concretely, imply that the preimage lookup extrinsic must be folded into state before the availability extrinsic may be fully processed and accumulation of work happen.
+The only synchronous entanglements are visible through the intermediate components superscripted with a dagger and defined in equations \ref{eq:betadagger}, \ref{eq:accountspostpreimage} and \ref{eq:rhoddagger}. The latter two mark a merge and join in the dependency graph and, concretely, imply that the preimage lookup extrinsic must be folded into state before the availability extrinsic may be fully processed and accumulation of work happen.
```
zdave-parity: "imply that the preimage lookup extrinsic must be folded into state before"... this is not accurate any more? Also `accountspostpreimage` is not superscripted with a dagger?

text/accumulation.tex
```diff
@@ -239,3 +217,32 @@ \subsection{Deferred Transfers and State Integration}
...
+    \end{aligned}
+  \right.
+\end{align}
+
+We define $\accountspostpreimage$ as the state after the integration of the preimages:
+\begin{align}
+  \accountspostpreimage = \accountspostxfer \text{ ex. } \forall \tup{s\ts\mathbf{p}} \in \xtpreimages : \left\{\,\begin{aligned}
+      \quad\accountspostpreimage[s]_\mathbf{p}[\mathcal{H}(\mathbf{p})] &= \mathbf{p} \\
+      \accountspostpreimage[s]_\mathbf{l}[\mathcal{H}(\mathbf{p}), |\mathbf{p}|] &= [\tau']
+    \end{aligned}\right.\!\!\!\!
```
zdave-parity, line 252: This is missing the "Any preimages whose integration became invalid during accumulation should be ignored but should NOT imply an invalid block" bit from #81.

text/header.tex
```diff
@@ -19,10 +19,13 @@ \section{The Header}\label{sec:header}
...
-The extrinsic hash is the hash of the block's extrinsic data. Given any block $\mathbf{B} = (\mathbf{H}, \mathbf{E})$, then formally:
-\begin{equation}
-  \mathbf{H}_x \in \H \,,\quad \mathbf{H}_x \equiv \mathcal{H}(\mathcal{E}(\mathbf{E}))
-\end{equation}
+The extrinsic hash is a Merkle commitment to the block's extrinsic data, taking care to allow for the possibility of reports to individually have their inclusion proven. Given any block $\mathbf{B} = (\mathbf{H}, \mathbf{E})$, then formally:
+\begin{align}
+  \mathbf{H}_x &\in \H \ ,\quad
+  \mathbf{H}_x \equiv \mathcal{H}(\se(\mathcal{H}^\#(\mathbf{a}))) \\
+  \where \mathbf{a} &= [\se_T(\xttickets), \se_P(\xtpreimages), g, \se_A(\xtassurances), \se_D(\xtdisputes)] \\
+  \also g &= \se(\mathcal{H}^\#(\se^\#_g(\xtguarantees)))
```
zdave-parity: This isn't quite compatible with JAM-SNP as currently specified, as that works with work-report hashes rather than guarantee hashes. I have two reasons for preferring WR hashes over guarantee hashes:

- It's possible for multiple guarantees to exist for the same WR. eg Different combinations of guarantor signatures are possible, and the timeslot can be trivially changed.
- WR hashes are used in judgments, and I think we may also need to use the get-a-work-report-given-a-hash protocol (CE 136) for handling negative judgments in the case where we weren't previously aware of the work-report being judged.

text/pvm_invocations.tex
```diff
@@ -753,8 +775,27 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
+  \texttt{void} = 22 \\
+  $g = 10$} &
+  $\begin{aligned}
+    \using [n, b, l] &= \registers_{7 \dots 10} \\
+    \using \mathbf{u} &= \begin{cases}
+      \mathbf{m}[n]_\mathbf{u} &\when n \in \keys{\mathbf{m}} \\
+      \error &\otherwise\\
+    \end{cases} \\
+    \using \mathbf{u}' &= \mathbf{u} \exc \begin{cases}
+      (\mathbf{u}'_\mathbf{A})_{\floor{\nicefrac{b}{\mathsf{Z}_P}}\dots\floor{\nicefrac{b+l-1}{\mathsf{Z}_P}} + 1} = [\none, \none, \dots]
```
zdave-parity: This should also be defined to zero the memory, otherwise it seems possible to revive old data using zero/poke with non-page-aligned addresses?

text/pvm.tex
```diff
@@ -637,9 +637,9 @@ \subsection{Standard Program Initialization}\label{sec:standardprograminit}
...
   &\using \mathcal{E}_3(|\mathbf{o}|) \concat \mathcal{E}_3(|\mathbf{w}|) \concat \mathcal{E}_2(z) \concat \mathcal{E}_3(s) \concat \mathbf{o} \concat \mathbf{w} \concat \mathcal{E}_4(|\mathbf{c}|) \concat \mathbf{c} = \mathbf{p}\\
-  &\mathsf{Z}_P = 2^{14}\ ,\quad\mathsf{Z}_Q = 2^{16}\ ,\quad\mathsf{Z}_I = 2^{24}\\
-  &\using \rnp{x \in \N} \equiv \mathsf{Z}_P\left\lceil \frac{x}{\mathsf{Z}_P} \right\rceil\quad,\qquad\rnq{x \in \N} \equiv \mathsf{Z}_Q\left\lceil \frac{x}{\mathsf{Z}_Q} \right\rceil\\
-  &5\mathsf{Z}_Q + Q(|\mathbf{o}|) + Q(|\mathbf{w}| + z\mathsf{Z}_P) + Q(s) + \mathsf{Z}_I \leq 2^{32}
+  &\mathsf{Z}_G = 2^{14}\ ,\quad\mathsf{Z}_Q = 2^{16}\ ,\quad\mathsf{Z}_I = 2^{24}\\
+  &\using \rnp{x \in \N} \equiv \mathsf{Z}_G\left\lceil \frac{x}{\mathsf{Z}_G} \right\rceil\quad,\qquad\rnq{x \in \N} \equiv \mathsf{Z}_Q\left\lceil \frac{x}{\mathsf{Z}_Q} \right\rceil\\
+  &5\mathsf{Z}_Q + Q(|\mathbf{o}|) + Q(|\mathbf{w}| + z\mathsf{Z}_G) + Q(s) + \mathsf{Z}_I \leq 2^{32}
 \end{align}
 Thus, if the above conditions cannot be satisfied with unique values, then the result is $\none$, otherwise it is a tuple of $\mathbf{c}$ as above and $\mem$, $\registers$ such that:
```
zdave-parity, line 705: Think this bit needs updating for the per-page access table?

text/header.tex
gavofyork: But this is a commitment to the *extrinsic*, not some edited components of the extrinsic. The extrinsic contains guarantees, not reports, and they're non-fungible.

Are you suggesting we instead have two-level hashing for guarantees in order that reports can be pre-distributed?

text/pvm.tex
gavofyork: Why? All of the memory segment sizes are already page aligned.

text/header.tex
zdave-parity: The idea was to pre-distribute reports yes, as this needs to happen anyway to make the reports available to block builders. So when distributing a new block, instead of sending the full block, you send a "light" block with the work reports replaced by their hashes. With the extrinsic hash as you have defined it, it's not possible to verify that such a "light" block is correct without obtaining all the referenced work reports. If the extrinsic hash instead at some level contained the work report hashes then this would be possible. You could add an extra level of hashing, or you could move the guarantee hash "down" to just cover the work report, and include the rest of the guarantee bits inline in `g`.

zdave-parity: The idea was to pre-distribute guarantees, as this needs to happen anyway to make the guarantees available to block builders. When distributing a new block, instead of sending the full block, you send a "light" block with the _work reports_ replaced by their hashes. You then request any work reports you don't have using CE 136. You could of course have the "light" block replace things at the level of guarantees instead of work reports. I figured the work report level was better for the reasons given above. To allow this to work, you could add an extra level of hashing, or you could move the guarantee hash "down" to just cover the work report, and include the rest of the guarantee bits inline in `g`. Alternatively we could change SNP to have "light" blocks include guarantee hashes and add a separate protocol for requesting guarantees. I just don't want to do that without explicit agreement that this is the preferred way forward.

text/pvm.tex
zdave-parity: Maybe I'm reading it wrong but I think `\mem` is supposed to have type `\ram`? `\ram` is a pair of ([u8]_2^32, [{0,W,R}]_numpages)? You say in the "Virtual Machine and Gas" section that `\mem_i` implicitly means `(\mem_V)_i`, which doesn't work here. Probably instead of `\mem_i` here there should have been `(\mem_i, (\mem_A)_i)`? But with the access table change I think it should instead say `(\mem_i, (\mem_A)_{i/pagesize})` or something like that.

text/pvm_invocations.tex
```diff
@@ -785,6 +785,7 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
       \error &\otherwise\\
     \end{cases} \\
     \using \mathbf{u}' &= \mathbf{u} \exc \begin{cases}
+      (\mathbf{u}'_\mathbf{V})_{b\dots+l} = [0, 0, \dots] \\
```
zdave-parity: Need to expand b..l out to page boundaries?

text/header.tex
gavofyork: ok - please see change

text/intro.tex
```diff
@@ -27,19 +27,20 @@ \subsection{Driving Factors}
...
 
-\subsection{Scaling under Size-Synchrony Antagonism}
+\subsection{Scaling under Size-Coherency Antagonism}
 
-Size-synchrony antagonism is a simple principle implying that as the state-space of information systems grow, then the system necessarily becomes less synchronous. The argument goes:
+Size-coherency antagonism is a simple principle implying that as the state-space of information systems grow, then the system necessarily becomes less coherent. It is a direct implication of of principle that causality is limited by speed. The maximum speed allowed by physics is $C$ the speed of light in a vacuum, however other information systems may have lower bounds: In biological system this is largely determined by various chemical processes whereas in electronic systems is it determined by the speed of electrons in various substances. Distributed software systems will tend to have much lower bounds still, being dependent on a substrate of software, hardware and packet-switched networks of varying reliability.
```
zdave-parity: of of

text/intro.tex
```diff
@@ -27,19 +27,20 @@ \subsection{Driving Factors}
...
+Setting the question of overall security aside for a moment, we can manage incoherence by fragmenting the system into causally-independent subsystems, each of which is small enough to be coherent. In a resource-rich environment, a bacterium may split into two rather than growing to double its size. This pattern is rather a crude means of dealing with incoherency under growth: intra-system processing has low size and high synchrony, inter-system processing supports higher overall sizes but with only low synchrony. It is the principle behind meta-networks such as Polkadot, Cosmos and the predominant vision of a scaled Ethereum (all to be discussed in depth shortly).
```
zdave-parity: Maybe s/synchrony/coherency/ here?

text/intro.tex
```diff
@@ -27,19 +27,20 @@ \subsection{Driving Factors}
...
+The present work explores a middle-ground in the antagonism, avoiding the persistent fragmentation of state-space of the system as with existing approaches. We do this by introducing a new model of computation which pipelines a highly scalable element to a highly synchronous element while retaining a storage facility which is \emph{mostly coherent}. Asynchrony is not avoided, but we do open the possibility for a greater degree of granularity over how it is traded against size. In particular fragmentation can be made ephemeral rather than persistent, drawing upon a coherent state and fragmenting it only for as long as it takes to execute any given piece of processing on it.
```
zdave-parity: And synchronous, asynchrony here?

text/pvm_invocations.tex
```diff
@@ -716,59 +716,52 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \texttt{peek} = 19 \\
   $g = 10$} &
   $\begin{aligned}
-    \using [n, a, b, l] &= \registers_{7 \dots 11} \\
+    \using [n, o, s, l] &= \registers_{7 \dots 11} \\
     \using \mathbf{s} &= \begin{cases}
       \none &\when n \not\in \keys{\mathbf{m}}\\
-      \error &\when \N_{b\dots+l} \not\in \mathbb{V}_{\mathbf{m}[n]_\mathbf{u}} \\
-      \mathbf{m}[n]_{\mathbf{u}_{b\dots+l}} &\otherwise
+      \mathbf{m}[n]_{\mathbf{u}_{s\dots+l}} &\when \N_{s\dots+l} \in \mathbb{V}_{\mathbf{m}[n]_\mathbf{u}} \wedge \N_{o\dots+l} \in \mathbb{V}^*_\memory \\
```
zdave-parity: For the latter condition seems like `\subseteq` should be used rather than `\in`?

text/pvm_invocations.tex
```diff
@@ -716,59 +716,52 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
     \end{cases} \\
-    \using \mathbf{s} &= \begin{cases}
-      \mem_{a\dots+l} &\when \N_{a\dots+l} \in \mathbb{V}_\mathbf{u}\\
-      \error &\otherwise
-    \end{cases}\\
     \using \mathbf{u}' &= \mathbf{u} \exc \begin{cases}
-      (\mathbf{u}'_\mathbf{V})_{b\dots+l} = \mathbf{s} \\
-      (\mathbf{u}'_\mathbf{A})_{\floor{\nicefrac{b}{\mathsf{Z}_P}}\dots\floor{\nicefrac{b+l-1}{\mathsf{Z}_P}} + 1} = [\mathrm{W}, \mathrm{W}, ...]
+      (\mathbf{u}'_\mathbf{V})_{p\mathsf{Z}_P\dots+c\mathsf{Z}_P} = [0, 0, \dots] \\
+      (\mathbf{u}'_\mathbf{A})_{p\dots c} = [\mathrm{W}, \mathrm{W}, \dots]
```
zdave-parity: Missing a + after `\dots` here.

text/pvm_invocations.tex
```diff
@@ -716,59 +716,52 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
     \end{cases}\\
     (\registers'_7, \mathbf{m}') &\equiv \begin{cases}
-      (\mathtt{OOB}, \mathbf{m}) &\when \mathbf{s} = \error \\
       (\mathtt{WHO}, \mathbf{m}) &\otherwhen \mathbf{u} = \error \\
```
zdave-parity: s/otherwhen/when/

text/pvm_invocations.tex
```diff
@@ -779,14 +772,14 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
       \error &\otherwise\\
     \end{cases} \\
     \using \mathbf{u}' &= \mathbf{u} \exc \begin{cases}
-      (\mathbf{u}'_\mathbf{V})_{b\dots+l} = [0, 0, \dots] \\
-      (\mathbf{u}'_\mathbf{A})_{\floor{\nicefrac{b}{\mathsf{Z}_P}}\dots\floor{\nicefrac{b+l-1}{\mathsf{Z}_P}} + 1} = [\none, \none, \dots]
+      (\mathbf{u}'_\mathbf{V})_{p\mathsf{Z}_P\dots+c\mathsf{Z}_P} = [0, 0, \dots] \\
+      (\mathbf{u}'_\mathbf{A})_{p\dots c} = [\none, \none, \dots]
     \end{cases}\\
     (\registers'_7, \mathbf{m}') &\equiv \begin{cases}
       (\mathtt{WHO}, \mathbf{m}) &\otherwhen \mathbf{u} = \error \\
```
zdave-parity: s/otherwhen/when/ here also, sorry didn't spot this before.

text/pvm_invocations.tex
```diff
@@ -779,14 +772,14 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
+    \using [n, p, c] &= \registers_{7 \dots 10} \\
     \using \mathbf{u} &= \begin{cases}
       \mathbf{m}[n]_\mathbf{u} &\when n \in \keys{\mathbf{m}} \\
       \error &\otherwise\\
     \end{cases} \\
     \using \mathbf{u}' &= \mathbf{u} \exc \begin{cases}
-      (\mathbf{u}'_\mathbf{V})_{b\dots+l} = [0, 0, \dots] \\
-      (\mathbf{u}'_\mathbf{A})_{\floor{\nicefrac{b}{\mathsf{Z}_P}}\dots\floor{\nicefrac{b+l-1}{\mathsf{Z}_P}} + 1} = [\none, \none, \dots]
+      (\mathbf{u}'_\mathbf{V})_{p\mathsf{Z}_P\dots+c\mathsf{Z}_P} = [0, 0, \dots] \\
+      (\mathbf{u}'_\mathbf{A})_{p\dots c} = [\none, \none, \dots]
```
zdave-parity: Missing a + after `\dots` here too.

text/pvm_invocations.tex
zdave-parity: Actually for both conditions I think you want subset.

text/header.tex
zdave-parity: LGTM, thanks :ok_hand:

text/pvm_invocations.tex
gavofyork: This was actually a problem in several places...

gavofyork:
@koute could you check this most recent commit? Also in `mul_32` and `add_32` do we sign-extend the result or leave as-is? I presume in `sub_32` we sign-extend.

text/pvm.tex
```diff
@@ -570,29 +570,31 @@ \subsubsection{Instructions with Arguments of Three Registers}
   \midrule
   \endhead
   170&\token{add\_32}&0&$\reg'_D = \sext_4((\reg_A + \reg_B) \bmod 2^{32})$\\ \mrule
-  171&\token{sub\_32}&0&$\reg'_D = \sext_4((\reg_A + 2^{32} - \reg_B) \bmod 2^{32})$\\ \mrule
+  171&\token{sub\_32}&0&$\reg'_D = \sext_4((\reg_A + 2^{32} - (\reg_B \bmod 2^{32})) \bmod 2^{32})$\\ \mrule
   172&\token{mul\_32}&0&$\reg'_D = \sext_4((\reg_A \cdot \reg_B) \bmod 2^{32})$\\ \mrule
   173&\token{div\_u\_32}&0&$\reg'_D = \begin{cases}
     2^{64} - 1 &\when \reg_B = 0\\
-    \sext_4(\floor{\reg_A \div \reg_B}) &\otherwise
+    \floor{(\reg_A \bmod 2^{32}) \div (\reg_B \bmod 2^{32})} &\otherwise
```
koute: The `\reg_B = 0` here still needs a modulo, otherwise we'll get division by zero if e.g. `ω_B = 0x00000001_00000000`

text/pvm.tex
```diff
@@ -570,29 +570,31 @@ \subsubsection{Instructions with Arguments of Three Registers}
...
-    \unsigned{\signed{\reg_A} \bmod \signed{\reg_B}} &\otherwise
+    \unsigned{a} &\when b = 0 \\
+    0 &\when a = -2^{31} \wedge b = -1 \\
+    \unsigned{a \bmod b} &\otherwise \\[2pt]
+    \multicolumn{2}{l}{\quad \where a = \signedn{4}{\reg_A \bmod 2^{32}}\,,\ b = \signedn{4}{\reg_B \bmod 2^{32}}}\\
   \end{cases}$\\ \mrule
   177&\token{shlo\_l\_32}&0&$\reg'_D = (\reg_A \cdot 2^{\reg_B \bmod 32}) \bmod 2^{32}$\\ \mrule
-  178&\token{shlo\_r\_32}&0&$\reg'_D = \floor{\reg_A \div 2^{\reg_B \bmod 32}}$\\ \mrule
-  179&\token{shar\_r\_32}&0&$\reg'_D = \unsigned{\floor{\signed{\reg_A} \div 2^{\reg_B \bmod 32}}}$\\ \mrule
+  178&\token{shlo\_r\_32}&0&$\reg'_D = \floor{(\reg_A \bmod 2^{32}) \div 2^{\reg_B \bmod 32}}$\\ \mrule
```
koute: The `shlo_l_32` and `shlo_r_32` are still missing `\sext_4`.

text/overview.tex
```diff
@@ -166,20 +166,24 @@ \subsection{The Virtual Machine and Gas}\label{sec:virtualmachineandgas}
...
+  \mathbb{M} &\equiv \tuple{
+    \isa{\mathbf{V}}{\Y_{2^{32}}},
+    \isa{\mathbf{A}}{\seq{\{\text{W}, \text{R}, \none\}}_p}
+  }\,,\ p = \frac{2^{32}}{\mathsf{Z}_P}\\
+  \mathsf{Z}_P &= 2^{12}
 \end{align}
 
-The \textsc{pvm} assumes a simple pageable \textsc{ram} of 32-bit addressable octets where each octet may be either immutable, mutable or inaccessible. The \textsc{ram} definition $\mathbb{M}$ includes two components: a value $\mathbf{V}$ and access $\mathbf{A}$. If the component is unspecified while being subscripted then the value component may be assumed. Within the context of the virtual machine, $\memory \in \mathbb{M}$ is typically used to denote \textsc{ram}.
+The \textsc{pvm} assumes a simple pageable \textsc{ram} of 32-bit addressable octets situated in pages of $2^p = 4096$ octets where each page may be either immutable, mutable or inaccessible. The \textsc{ram} definition $\mathbb{M}$ includes two components: a value $\mathbf{V}$ and access $\mathbf{A}$. If the component is unspecified while being subscripted then the value component may be assumed. Within the context of the virtual machine, $\memory \in \mathbb{M}$ is typically used to denote \textsc{ram}.
```
koute, line 183: Just below this text we have (I can't put my comment directly there as GH disallows that for parts of the file that were not changed):

> A page fault (attempt to access some address in \textsc{ram} which is not accessible), \\$\fault\\$. This includes the address at fault.

It'd probably be better/simpler to give out the address of the *page* that is at fault (or the page index, if you prefer), and not the exact address that was accessed (considering the program will have to page in the whole page anyway)? Implementation-wise we *can* figure out the exact address that was accessed, but that could have some portability issues on different/future OSes or hardware architectures.

text/pvm_invocations.tex
```diff
@@ -732,34 +734,67 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
+  \makecell*[l]{
+  $\Omega_Z(\gascounter, \registers, \memory, (\mathbf{m}, \mathbf{e}))$ \\
+  \texttt{zero} = 21 \\
+  $g = 10$} &
+  $\begin{aligned}
+    \using [n, p, c] &= \registers_{7 \dots 10} \\
     \using \mathbf{u} &= \begin{cases}
       \mathbf{m}[n]_\mathbf{u} &\when n \in \keys{\mathbf{m}} \\
       \error &\otherwise\\
     \end{cases} \\
```
koute, line 751: Zero should return an error if the program tries to allocate any pages in the lower 64k of address space/the lower 16 pages of address space.

koute:
Also, the host call result constants (NONE, WHAT, OOB, etc.) are currently defined as `2^32 - n` - we should probably change them to `2^64 - n`?

koute (approved):

PR #151, opened 2024-11-22
Allow 8 octets input signed extension function (64bit)

emielsebastiaan:
# Suggested changes (32bit -> 64bit)

1. In GP-0.5.0-eq:A.11: Allow 8 input octets for the signed extension function as a consequence of the change to a 64bit architecture. Direct evidence of this requirement in instructions 141 & 142.
2. In GP-0.5.0-eq:A.20: add mod2^32 to instruction 72.


koute:
The limit of immediates being at most 4 bytes is a deliberate design decision. We don't want 8 byte immediates; it just makes recompilation into native code slower and more complex without any real benefit.

koute:
8 byte *physical* (as encoded in the code blob) immediates are not allowed (except the `load_imm_64` instruction). The immediate value (as encoded in the code blob) can be at most 4 bytes, but if necessary it is sign extended to full 64-bit. The difference between `store_imm_u32` and `store_imm_u64` is that `store_imm_u32` will write the lower four bytes of the sign extended immediate to memory, and `store_imm_u64` will write the full 8 bytes of the sign extended immediate.

emielsebastiaan:
Thank you for the comments @koute 
I changed this PR. 
Two small suggestions remain.

gavofyork (approved):

PR #152, opened 2024-11-22
Fix PVM page size constant usage

0xjunha:
* Refer to the correct page size constant in the overview.
* Use the initialization page size Z_G instead of Z_P in the standard initialization decoder function Y.


gavofyork:
The main change of this is now unneeded, but the first one is still good - thanks!

https://github.com/gavofyork/graypaper/pull/159

PR #154, opened 2024-11-25
EpochMarker use η0 and η1 instead of η1' and η2'

sourabhniyogi:
* η0 and η1 is equivalent to η1' and η2'  based on [this](https://graypaper.fluffylabs.dev/#/911af30/0e57020e5702)
* This keeps the dependency graph [here](https://graypaper.fluffylabs.dev/#/911af30/098001098001) clean and easy to reason about.


gilescope:
Thank you. I was wondering if there was some subtle reason why this simplification wasn't done. Definitely simpler to understand this way.

PR #160, opened 2024-12-06
Corrections on 0.5.2

gavofyork:
- [Definitions: Update gas limits so that gas = nanoseconds](https://github.com/gavofyork/graypaper/pull/160/commits/83f7ad5601df239040c7caa4c0b52b590205f997)
- [WP&WRs: Badly reported exports count doesn't invalidate WP](https://github.com/gavofyork/graypaper/pull/160/commits/184515d80354f96f0410b48cea950ed7c9e4f03f)
- [ReportingAssurance: Active removal of expired reports](https://github.com/gavofyork/graypaper/pull/160/commits/3e99734afc3770b145539f20b1cfa2939bbecd80)

### DIFF
--- a/text/reporting_assurance.tex
+++ b/text/reporting_assurance.tex
@@ -82,10 +82,10 @@ Next, the hash of the payload ($l$) within the work item which was executed in t
 
 Finally, there is the output or error of the execution of the code $\mathbf{o}$, which may be either an octet sequence in case it was successful, or a member of the set $\mathbb{J}$, if not. This latter set is defined as the set of possible errors, formally:
 \begin{equation}
-  \mathbb{J} \in \{ \oog, \panic, \token{BAD}, \token{BIG} \}
+  \mathbb{J} \in \{ \oog, \panic, \badexports, \token{BAD}, \token{BIG} \}
 \end{equation}
 
-The first two are special values concerning execution of the virtual machine, $\oog$ denoting an out-of-gas error and $\panic$ denoting an unexpected program termination. Of the remaining two, the first indicates that the service's code was not available for lookup in state at the posterior state of the lookup-anchor block. The second indicates that the code was available but was beyond the maximum size allowed $\mathsf{W}_C$.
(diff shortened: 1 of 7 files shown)


text/reporting_assurance.tex
```diff
@@ -124,10 +124,10 @@ \subsubsection{The Assurances Extrinsic}
   &\mathsf{X}_A \equiv \token{\$jam\_available}
 \end{align}
 
-A bit may only be set if the corresponding core has a report pending availability on it which has not timed out:
+A bit may only be set if the corresponding core has a report pending availability on it:
 \begin{align}
   &\forall a \in \xtassurances, c \in \N_\mathsf{C} :\\
-  &\quad a_f[c] \Rightarrow \rho^\dagger[c] \ne \none \wedge \mathbf{H}_t \le \rho^\dagger[c]_t + \mathsf{U}
+  &\quad a_f[c] \Rightarrow \rho^\dagger[c] \ne \none
```
davxy: Perhaps we can eliminate the second condition (timeout) from 11.30. 
By this stage, it should already have been reset to the empty set as per 11.18 ($\mathbf{H}_t \ge \rho^\dagger[c]_t + \mathsf{U}$).

screenshot-2024-12-08-11-15-55

text/reporting_assurance.tex
```diff
@@ -124,10 +124,10 @@ \subsubsection{The Assurances Extrinsic}
   &\mathsf{X}_A \equiv \token{\$jam\_available}
 \end{align}
 
-A bit may only be set if the corresponding core has a report pending availability on it which has not timed out:
+A bit may only be set if the corresponding core has a report pending availability on it:
```
davxy, line 127: So, according to this change, now a bit can be set even if the pending report has timed out. Thus , if we achieve an availability supermajority for a timed-out report (according to 11.17) it would still be included in **W**.

screenshot-2024-12-08-11-26-15

This means it will still be considered for accumulation in the next pipeline stage.

Essentially, we're now a bit more lenient with timeouts. In other words, if it's timed out but still available, we no longer consider it an issue.

Given this explicit change, I assume this is an intentional design choice. However, I just wanted to double-check to be sure

gavofyork: Yes - if a report is made available on the exact timeout block, we prefer to accumulate than to drop.

PR #180, opened 2025-01-03
fix missed is_authorized p_c to p_u

qiweiii:
it's probably `p_u`, there is no `c` in work package definition now


gavofyork:
image

See 14.9

PR #196, opened 2025-01-20
chore(overview): add tau to the dependencies of disputes

clearloop:
tau(time slot) is used for checking the age of verdicts in the state transition of psi(disputes), see also graypaper 10.2 >> (10.2):

https://github.com/gavofyork/graypaper/blob/16d9709d14a2afaeaefe9fbdca07d58cfcdc88c9/text/judgments.tex#L41


clearloop:
closing since the dependency graph is a bit more complicated than I thought

PR #199, opened 2025-01-28
Remove OOB and panic on invalid mem access

gavofyork:


text/pvm_invocations.tex
```diff
@@ -769,10 +765,10 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
       \error &\otherwise
     \end{cases}\\
-    (\registers'_7, \mem') &\equiv \begin{cases}
-      (\mathtt{OOB}, \mem) &\when \mathbf{s} = \error \\
-      (\mathtt{WHO}, \mem) &\when \mathbf{s} = \none \\
-      (\mathtt{OK}, \mem') \ \where \mem' = \mem \exc \memwr_{o\dots+z} = \mathbf{s} &\otherwise
+    (\execst', \registers'_7, \mem') &\equiv \begin{cases}
+      (\panic, \registers_7, \mem) &\when \mathbf{s} = \error \\
+      (\continue, \mathtt{WHO}, \mem) &\when \mathbf{s} = \none \\
+      (\continue, \mathtt{OK}, \mem') \ \where \mem' = \mem \exc \memwr_{o\dots+z} = \mathbf{s} &\otherwise
```
koute: For `peek` and `poke` the `OOB` is still useful if the given region of memory of the *inner* VM is not accessible.

Panicking in case the pointer to the *outer* VM's memory is invalid makes sense since making sure that's the case is essentially "free" - in their services people will just allocate a buffer and pass a pointer to this buffer, and they have to go out of their way to pass something invalid here. But this is not necessarily true for inner VMs.

For example, let's say the service is a smart contract host and runs a smart contract in an inner VM. Then the smart contract triggers a host call asking the outer VM to read some data from a pointer controlled by the inner VM. Does that pointer point somewhere valid within the inner VM? It may or may not.

So in this case having OOB would be useful for the outer VM to be able to easily abort the execution of the inner VM without panicking itself.

Technically the outer VM could track which pages are allocated and which aren't (since it responds to all page faults), but that sounds like a huge pain, especially in cases where the inner VM won't be suspended/resumed across blocks so the outer VM isn't forced to track which pages are used.

gavofyork: Yes fair enough

koute (approved):

PR #227, opened 2025-02-05
PVM: Signed division should always round to zero

koute:
As reported in: https://github.com/w3f/jamtestvectors/pull/3#issuecomment-2632729448

Anywhere we divide a potentially negative number and we want a non-floating-point result we need to round towards zero instead of flooring (the fractional part of the result should always be discarded).

### DIFF
--- a/text/pvm.tex
+++ b/text/pvm.tex
@@ -43,6 +43,7 @@
 \newcommand*{\immed}{\nu}
 \newcommand*{\deblob}{\text{deblob}}
 \newcommand*{\smod}{\text{smod}}
+\newcommand*{\rtz}{\text{rtz}}
 
 We declare the general \textsc{pvm} function $\Psi$. We assume a single-step invocation function define $\Psi_1$ and define the full \textsc{pvm} recursively as a sequence of such mutations up until the single-step mutation results in a halting condition. We additionally define the function $\deblob$ which extracts the instruction data, opcode bitmask and dynamic jump table from a program blob:
 \begin{align}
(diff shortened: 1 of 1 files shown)


emielsebastiaan:
This is likely an incomplete solution. We implemented these changes and now test vectors for **all** `shar*` (Shift Arithmetic Right) testvectors fail.
I.e., instructions: 140, 146, 153, 157, 199 & 209

# Example: 
- `shar_r_64` (209)
- Testvector: shift_arithmic_right_64.json

## Input
ω_A (register: 7) = 9223372036854775925
ω_B (register: 8) = 3
ω_D (register: 9) = Destination register for outcome

## Calculations
Z_8(ω_A) = -9223372036854775691
2**(ω_B % 64) = 8
Division: -9223372036854775691 / 8 = -1152921504606846961,375
Floor/Ceil: rtz(-9223372036854775691 / 8) = rtz(-1152921504606846961,375) = -1152921504606846961
Calculated output: ω_D (register: 9) = Z_inv_8(-1152921504606846961) = 17293822569102704655

## Expected Output Test Vectors
ω_D= 17293822569102704654 (!= Calculated output: 17293822569102704655)

# Suggested follow-up
- Possible GP definition issue with current `shar*` instructions: 140, 146, 153, 157, 199 & 209 (or less likely test vectors are incorrect).
- Unconfirmed issues since no testvectors yet exist with instructions `mul_upper_s_s` (213) & `mul_upper_s_u` (215).

koute:
Hm, at least with arithmetic shifts you're correct. I'll double check this. Thanks.

koute:
Okay, I double checked, and I indeed had a brain fart here. Apologies. Only divisions should have this; thanks @emielsebastiaan.

emielsebastiaan:
Yes confirmed... in all those other modifications you made ealier (140, 146, 153, 157, 199, 209, 213 & 215) the division only protected against overflow. All good now! thanks

PR #248, opened 2025-02-24
Section 11: beta to beta_dagga

yoyo2325:
When working on extrinsics guarantee-related programs, I found that the beta state has two reference targets. However, the beta dagga should actually be applicable to all Section 11 extrinsics guarantee-related verification.

ref beta dagga here
https://graypaper.fluffylabs.dev/#/5f542d7/153201153301
but ref beta here
https://graypaper.fluffylabs.dev/#/5f542d7/15b10115ee01


gavofyork:
It is, but I'm not sure why we would use a later state item over an earlier one, since it suggests an dependency which does not exist.

yoyo2325:
But if we adopt only one option that can reduce complexity and the chance of misunderstandings

yoyo2325:
In the rest of the sections, there will be a clear sequence, except for this one.

gavofyork:
Closing since there's a general convention to use the simplest formulation. $\beta^\dagger$ is a more complex formulation than $\beta$. I see a greater chance of misunderstandings arise from involving terms which are strictly more complex than are needed.

PR #253, opened 2025-02-26
Update stf dependency order for β′

jaymansfield:
β′ requires the commitment map which doesn't exist at it's current point in the dependency graph.

This PR corrects the order.


sourabhniyogi:
image

This change is a clear improvement because the C of 4.17 is actually dependent on 4.7 and cannot be parallelized.

PR #260, opened 2025-03-02
Definitions: Correct constant `D` value

0xjunha:
Update constant `D` from 28,000 to 4,800 to align with [this commit](https://github.com/gavofyork/graypaper/commit/19ebc483086d40d37531c0a43d51a49bcc61d859) and [its usage in the PVM invocations section](https://github.com/gavofyork/graypaper/blob/main/text/pvm_invocations.tex#L58)


gavofyork:
This is probably not the correct fix. Issue added https://github.com/gavofyork/graypaper/issues/268

PR #261, opened 2025-03-02
Use explicit tuple keys for preimage lookup metadata dictionaries

0xjunha:
# Summary
When accessing entries of service preimage lookup dictionary (`l`), omitting parentheses from tuple keys is understandable given the clear definition of the dictionary. However, suggesting explicitly using tuple keys for consistency, as seen [here](https://github.com/gavofyork/graypaper/blob/main/text/accumulation.tex#L236) and [here](https://github.com/gavofyork/graypaper/blob/main/text/pvm_invocations.tex#L582-L583).

# Updates

### Historical Lookup Function
(prev)
prev_lookup_function

(updated)
lookup_function

### Preimages Integration STF
(prev)
prev_preimages_stf

(updated)
preimages_stf

### `eject` host function
(prev)
prev_host_eject

(updated)
host_eject

### `query` host function
(prev)
prev_host_query

(updated)
host_query

### `forget` host function
(prev)
prev_host_forget

(updated)
host_forget

### DIFF
--- a/text/accounts.tex
+++ b/text/accounts.tex
@@ -106,7 +106,7 @@ The historical lookup function $\Lambda$ may now be defined as:
   \begin{aligned}\label{eq:historicallookup}
     &\Lambda\colon (\mathbb{A}, \N_T, \H) \to \Y\bm{?} \\
     &\Lambda(\mathbf{a}, t, h) \equiv \begin{cases}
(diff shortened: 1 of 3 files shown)


gavofyork:
I prefer clarity over "machine readability". If it really bothers you I'd accept a PR explicitly stating in the notation section that sequence subscripting may, if done with a tuple, without the tuple's parentheses.

0xjunha:
Got it, for consistency, would it be better to remove parentheses from [here](https://github.com/gavofyork/graypaper/blob/d3ef7fc1d8dba223d99db86af806071f00ccc951/text/accumulation.tex#L249) and [here](https://github.com/gavofyork/graypaper/blob/d3ef7fc1d8dba223d99db86af806071f00ccc951/text/pvm_invocations.tex#L582-L583) instead?

PR #282, opened 2025-03-10
Include ed25519 keys in the epoch marker

arkpar:
This is useful for GRANDPA and calculating assurer sets for past epochs.

image

image


arkpar:
cc @davxy @zdave-parity

davxy (approved):
Just for my understanding of the actual added value of this info .

IIUI, the current information within the epoch marker is fundamental to help light clients validate blocks (full nodes can extract the same information from the state).

> This is useful for GRANDPA and calculating assurer sets for past epochs.

Is to help light clients follow GRANDPA protocol? Or this is just to prevent the annoying task of having to access the state at block.header.parent to get this info?

arkpar:
> Is to help light clients follow GRANDPA protocol? Or this is just to prevent the annoying task of having to access the state at block.header.parent to get this info?

w.r.t GRANDPA this can be used to follow/prove finality without followin the protocol. We might want to generate proofs of finality changes for blocks we don't have the state for e.g. for warp sync.

PR #290, opened 2025-03-14
Change B_L to B_O

DrEverr:
Renamed the constant from `B_L` to `B_O`.
`B_O` implies 'Octet'


gavofyork:
The L in B_L implies "length".

PR #291, opened 2025-03-15
WPs&WRs: SR lookup key should be work package hash

0xjunha:
SR lookup keys should be work package hash (h with boxplus tag), not the SR.


0xjunha:
After reviewing more closely, I've decided to close the PR after realizing the untagged `h` indicates the work package hash with no misleading.

PR #295, opened 2025-03-18
Explicitly serialize the work report before hashing

greywolve:


gavofyork:
It is currently unambiguous and in line with the definition in 3.8.1:

> The inputs of a hash function should be expected to be passed through our serialization codec E to yield an octet sequence to which the cryptography may be applied. 

I take the convention *not* to add functions unnecessarily in order to keep things concise and aid clarity for human readers, rather than a hypothetical machine with some sort of limited grammatical ability.

PR #296, opened 2025-03-19
Fix : service-id serial subscript

TwEricShen:
Since service code-hash is a 32-octet value,  it should be compared with E_4(service-id), rather than E_32(service-id), which is an 8*32-octet value.

link : https://graypaper.fluffylabs.dev/#/85129da/32fb0132fb01?v=0.6.3


gavofyork:
E_{4} takes a number and results in a sequence of 4 octets.

E_{32} takes a number and results in a sequence of 32 octets.

PR #304, opened 2025-03-24
Serializiation: Operand tuples: remove variable size

daiagi:
Variable size is not needed for the encoding of the **d** element as it is an octet sequence and fully determined by the `0` prefix


gavofyork:
It is still of arbitrary length and thus needs the prefix for proper decoding.

PR #305, opened 2025-03-24
Serialization: WorkReport: gas is 8 bytes

daiagi:


gavofyork:
This field is a statistic and intended to be variable width.

PR #313, opened 2025-03-28
cleanup use of accumulate host function wrapper

0trust3r:
flushing S back to partial state is moot for read-only host functions

this pr removes the use of the host function `G` wrapper from the **read, info, & lookup** host function collection for accumulation

https://graypaper.fluffylabs.dev/#/68eaa1f/2ec7012ec701?v=0.6.4

https://graypaper.fluffylabs.dev/#/68eaa1f/2ebb022ebb02?v=0.6.4

### DIFF
--- a/text/pvm_invocations.tex
+++ b/text/pvm_invocations.tex
@@ -145,11 +145,11 @@ We define $\Psi_A$, the Accumulation invocation function as:
     &\qquad\where i = \text{check}((\de_4(\mathcal{H}(\se(s, \eta'_0, \mathbf{H}_t))) \bmod (2^{32}-2^9)) + 2^8) \\
   \end{aligned}\right.\\
   F \in \Omega\ang{(\mathbf{X}, \mathbf{X})} &\colon (n, \gascounter, \registers, \memory, (\mathbf{x}, \mathbf{y})) \mapsto \begin{cases}
-    G(\Omega_R(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{read} \\
+    \Omega_R(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y}) &\when n = \mathtt{read} \\
     G(\Omega_W(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{write} \\
-    G(\Omega_L(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{lookup} \\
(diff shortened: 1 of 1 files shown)


0trust3r:
see [this post in the GP matrix](https://matrix.to/#/!ddsEwXlCWnreEGuqXZ:polkadot.io/$S7v9E-i18MuU_VzMQGeRkoxPd7NRnqYcmtBqKr2W6vM?via=polkadot.io&via=matrix.org&via=parity.io) for context

text/pvm_invocations.tex
```diff
@@ -145,11 +145,11 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
...
   \end{aligned}\right.\\
   F \in \Omega\ang{(\mathbf{X}, \mathbf{X})} &\colon (n, \gascounter, \registers, \memory, (\mathbf{x}, \mathbf{y})) \mapsto \begin{cases}
-    G(\Omega_R(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{read} \\
+    \Omega_R(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y}) &\when n = \mathtt{read} \\
     G(\Omega_W(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{write} \\
-    G(\Omega_L(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{lookup} \\
+    \Omega_L(\gascounter, \registers, \memory, \mathbf{s}, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y}) &\when n = \mathtt{lookup} \\
     \Omega_G(\gascounter, \registers, \memory, (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{gas} \\
-    G(\Omega_I(\gascounter, \registers, \memory, \mathbf{x}_s, \mathbf{d}), (\mathbf{x}, \mathbf{y})) &\when n = \mathtt{info} \\
```
gavofyork, line 152: This change is not syntactically correct. You can't just write two terms with a comma between them - it's meaningless.

gavofyork (changes requested):
Invalid syntax, and as I stated in the chat I'm not at all sure that the change you want to make is possible without over-complicating things.

text/pvm_invocations.tex
0trust3r: good catch, fixed in [548b309](https://github.com/gavofyork/graypaper/pull/313/commits/548b309904b1a6e0d067896162c6d008cf3c1649)

gavofyork: Again this is not correct. $\Omega\langle(\mathbf{X}, \mathbf{X})\rangle$ must result in a tuple as defined in A.35; in this case, the final tuple element must itself be the pair which makes up the Accumulation context. $\Omega_G$ is defined as being a pass-through for any context argument given and so can work without any additional wrangling. The other four cannot. That's pretty much the point of $G$.

0trust3r: I see your point now about wrangling, good catch. Your earlier comment about this change potentially over-complicating things makes sense now. I'll go ahead and close this pr. Thanks for following up on this thread.

PR #319, opened 2025-03-31
work report description

0trust3r:
recently there was a gas-related field added to the work report tuple

this pr updates the tuple description paragraph to include this field. **I also removed the results length description for sake of brevity/cleanliness, since it can be pretty easily gathered from the tuple field subscript itself that the length is 1..I** Happy to adjust here if needed.

*I'm assuming this newly-added field represents the gas consumed during refinement, based on the other changes from https://github.com/gavofyork/graypaper/pull/285*

### DIFF
--- a/text/reporting_assurance.tex
+++ b/text/reporting_assurance.tex
@@ -21,7 +21,7 @@ The state of the reporting and availability portion of the protocol is largely c
 As usual, intermediate and posterior values ($\rho^\dagger$, $\rho^\ddagger$, $\rho'$) are held under the same constraints as the prior value.
 
 \subsubsection{Work Report}\label{sec:workreport}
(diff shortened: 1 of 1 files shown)


0xjunha:
I agree that clarifying the `g` of work report in the description will be helpful.
However, this field should be gas used in `is_authorized` rather than refine:
https://github.com/gavofyork/graypaper/blob/c7f4b98cc6e9cc3b553504830269ceec197178b3/text/work_packages_and_reports.tex#L179
https://github.com/gavofyork/graypaper/blob/c7f4b98cc6e9cc3b553504830269ceec197178b3/text/work_packages_and_reports.tex#L188

Also [macro {\wr¬authgasused}](https://github.com/gavofyork/graypaper/blob/c7f4b98cc6e9cc3b553504830269ceec197178b3/text/reporting_assurance.tex#L35) is used.

0trust3r:
> Contributor
> ###

Thanks for the feedback! Nice catch on the invocation context- fixed in [2abea7a](https://github.com/gavofyork/graypaper/pull/319/commits/2abea7adaba9bca1b89020871e60e7b6d583a349)

Can you clarify your comment regarding [macro {\wr¬authgasused}](https://github.com/gavofyork/graypaper/blob/c7f4b98cc6e9cc3b553504830269ceec197178b3/text/reporting_assurance.tex#L35)? I didn't change this in my pr, are you requesting a change here?

0xjunha:
I just mentioned it as another evidence for the field being is_authorized gas usage - since its macro name is "authgasused" :)

PR #340, opened 2025-04-22
missing new type in parallelized accumulation

danicuki:
obs: I assumed that `p*` is a typo and changed to `p`. If this is not correct, maybe there is something missing to define what `p*` is.


danicuki:
I also noticed that ΨA does not return **p**, so Formula 12.20 might have something missing?
Screenshot 2025-04-23 at 09 32 18

PR #353, opened 2025-05-01
Set last byte to 0 in service storage key construction

davxy:
The last byte is unused in practice and can be safely dropped, which simplifies the process of recovering the original storage key from a trie key (e.g. during test vectors construction or state dumps)

To maintain consistency and convenience - i.e. to keep the 32-byte hashes domain - I propose to change the serialization rule for service keys where the last byte is just set to `0`. This makes the "full" storage key recoverable from the trie key.

Screenshot_2025-04-30_17-37-27

A non-zero value can also lead to confusion https://github.com/davxy/jam-test-vectors/pull/45#issuecomment-2840337660

This change does not impact security in any way, as the last byte is already dropped during Merkelization.

### DIFF
--- a/text/merklization.tex
+++ b/text/merklization.tex
@@ -12,7 +12,7 @@ We define the state-key constructor functions $C$ as:
     \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y} &\to \H \\
     i \in \N_{2^8} &\mapsto [i, 0, 0, \dots] \\
     (i, s \in \N_S) &\mapsto [i, n_0, 0, n_1, 0, n_2, 0, n_3, 0, 0, \dots]\ \where n = \se_4(s) \\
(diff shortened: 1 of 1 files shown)


davxy:
As discussed, a viable alternative is to directly define the storage key as a 31 bytes array

sourabhniyogi:
> As discussed, a viable alternative is to directly define the storage key as a 31 bytes array

For GP, this is the right thing to do given [D.2.1](https://graypaper.fluffylabs.dev/#/5f542d7/392500392500?v=0.6.2) 

image

I believe it would be useful in Appendix D exposition+CE129 (whenever that is included in GP) for this "31 byte" type to be treated as _primary_ with its own set of possible storage keys within ${\cal Y}$.  The "First 31 bytes of key only" type disclaimers should disappear.

While I'm sure many implementations will have a special type like "Hash" for the extremely common 32-byte case, there is no reason why a special type for 31 byte would not also have its own type following this exposition (or hack the 32-byte case by adding / removing the "0") -- at the end of the day this should be an implementation specific concern though -- the above exposition change would make it clear.

davxy:
Potentially superseded by https://github.com/gavofyork/graypaper/pull/356

PR #356, opened 2025-05-01
Reduce the size of state key(s) to 31 octets

davxy:
Perhaps a less confusing alternative to https://github.com/gavofyork/graypaper/pull/353

### DIFF
--- a/text/merklization.tex
+++ b/text/merklization.tex
@@ -9,7 +9,7 @@ The serialization of state primarily involves placing all the various components
 We define the state-key constructor functions $C$ as:
 \begin{equation}
   C\colon\left\{\begin{aligned}
-    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y_{28}} &\to \Y_{31} \\
+    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y_{27}} &\to \Y_{31} \\
     i \in \N_{2^8} &\mapsto [i, 0, 0, \dots] \\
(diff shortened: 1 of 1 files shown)


text/merklization.tex
```diff
@@ -1,18 +1,18 @@
...
 \subsection{Serialization}
 
-The serialization of state primarily involves placing all the various components of $\sigma$ into a single mapping from 32-octet sequence \emph{state-keys} to octet sequences of indefinite length. The state-key is constructed from a hash component and a chapter component, equivalent to either the index of a state component or, in the case of the inner dictionaries of $\delta$, a service index.
+The serialization of state primarily involves placing all the various components of $\sigma$ into a single mapping from 31-octet sequence \emph{state-keys} to octet sequences of indefinite length. The state-key is constructed from a hash component and a chapter component, equivalent to either the index of a state component or, in the case of the inner dictionaries of $\delta$, a service index.
 
 We define the state-key constructor functions $C$ as:
 \begin{equation}
   C\colon\left\{\begin{aligned}
-    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y} &\to \H \\
+    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y_{27}} &\to \Y_{31} \\
```
kianenigma, line 12: Should this not be `(N_s, Y_31) -> Y_31`?

kianenigma: Note that further below, the input to `C` is a 4 byte service id, then a 4 byte value concatenated with a 27 byte hash. So the input to C is `(N_s, Y_31)`

PR #357, opened 2025-05-01
use big endian for instructions 104-107

ascrivener:


gavofyork:
Real problem, incorrect fix. Fixzed by https://github.com/gavofyork/graypaper/pull/358

PR #384, opened 2025-05-11
typo: refine-fetch should pass Null instead of H0

prasad-kumkar:
Since we are checking if `n` is None in fetch call as such:
image

image

it should be passed as None instead of H0 in B.6

image


gavofyork:
No - the zero entropy is there intentionally as a placeholder for some real entropy to be introduced in the 0.7 series.

PR #387, opened 2025-05-12
Various nits related to fetch

ascrivener:


text/pvm_invocations.tex
```diff
@@ -92,7 +92,7 @@ \subsection{Refine Invocation}\label{sec:refineinvocation}
   &F \in \Omega\ang{(\dict{\N}{\pvm}, \seq{\G})} \colon
     (n, \gascounter, \registers, \memory, (\mathbf{m}, \mathbf{e})) \mapsto \begin{cases}
       \Omega_G(\gascounter, \registers, \memory, (\mathbf{m}, \mathbf{e})) &\when n = \mathtt{gas} \\
-      \Omega_Y(\gascounter, \registers, \memory, p, \H_0, \mathbf{r}, i, \overline{\mathbf{i}}, \overline{\mathbf{x}}, \none, \none, (\mathbf{m}, \mathbf{e})) &\when n = \mathtt{fetch}\\
+      \Omega_Y(\gascounter, \registers, \memory, p, \H_0, \mathbf{r}, i, \overline{\mathbf{i}}, \overline{\mathbf{x}}, \none, \none) &\when n = \mathtt{fetch}\\
```
gavofyork: The context cannot trivially be removed since the formulation requires it to be passed through the specific host call and into the result of the invocation.

gavofyork (changes requested):
I'll accept the second set of changes, but the first set don't work.

text/pvm_invocations.tex
ascrivener: sounds good. Then we need to address that sometimes (m,e) is passed, sometimes (x,y). The context is also sometimes not passed at all (on-transfer), and always passed as the last component incorrectly (the host function definition has it as the 4th component). Maybe could use (...) a la in the gas host function input components

ascrivener: scratch that, I see it's been done in 0.6.7

gavofyork (approved):

PR #406, opened 2025-05-27
Merklisation: Optimize state-tree entropy

### LINKED ISSUE #403: Revisit the state hashing
At present, 9 bytes of entropy are dropped from all service storage keys on their way into the state Merkle root. This is probably right on the edge of security.

Better would be to redistribute the entropy between the initial dropping of 5 bytes and the secondary dropping of 4 bytes to give a final state tree key which retains 27 bytes of entropy from the service storage key.

gavofyork:
Fixes #403

### DIFF
--- a/text/definitions.tex
+++ b/text/definitions.tex
@@ -288,7 +288,7 @@ Here, the prime annotation indicates posterior state. Individual components may 
   \item[$\mathsf{U} = 5$] The period in timeslots after which reported but unavailable work may be replaced.
   \item[$\mathsf{V} = 1023$] The total number of validators.
   \item[$\mathsf{W}_A = 64,000$] The maximum size of is-authorized code in octets.
-  \item[$\mathsf{W}_B = 12\cdot2^{20}$] The maximum size of an encoded work-package together with its extrinsic data and import implications, in octets.
+  \item[$\mathsf{W}_B = 13,794,305$] The maximum size of an encoded work-package together with its extrinsic data and import implications, in octets.
   \item[$\mathsf{W}_C = 4,000,000$] The maximum size of service code in octets.
   \item[$\mathsf{W}_E = 684$] The basic size of erasure-coded pieces in octets. See equation \ref{eq:erasurecoding}.
   \item[$\mathsf{W}_G = \mathsf{W}_P\mathsf{W}_E = 4104$] The size of a segment in octets.
--- a/text/work_packages_and_reports.tex
+++ b/text/work_packages_and_reports.tex
@@ -101,7 +101,7 @@ We limit the total number of exported items to $\mathsf{W}_X = 3072$, the total 
 
 We make an assumption that the preimage to each extrinsic hash in each work-item is known by the guarantor. In general this data will be passed to the guarantor alongside the work-package.
 
-We limit the total size of the implied import and extrinsic items, together with all payloads, the authorizer configuration and the authorization token to 12\textsc{mb} in order to allow for around 2\textsc{mb}/s/core data throughput:
+We limit the total size of the auditable \emph{work-bundle}, containing the work-package, import and extrinsic items, together with all payloads, the authorizer configuration and the authorization token to around 13.6\textsc{mb}. This limit allows 2\textsc{mb}/s/core D$^{3}$L imports, and thus a full complement of 3,072 imports, assuming no extrinsics and a minimal work-package of 4\textsc{kb}:
 \begin{align}
   \label{eq:checkextractsize}
   &\begin{aligned}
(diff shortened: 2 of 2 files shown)


text/merklization.tex
```diff
@@ -41,13 +41,13 @@ \subsection{Serialization}
     &&C(15) &\mapsto \se([\var{\mathbf{i}} \mid \mathbf{i} \orderedin \accumulated]) \;, \\
     &&C(16) &\mapsto \se(\var{[(\se_4(s), \se(h)) \mid (s, h) \orderedin \lastaccout]}) \;, \\
     \forall (s \mapsto \mathbf{a}) \in \delta: &&C(255, s) &\mapsto \mathbf{a}_c \concat \se_8(\mathbf{a}_b, \mathbf{a}_g, \mathbf{a}_m, \mathbf{a}_o, \mathbf{a}_f) \concat \se_4(\mathbf{a}_i, \mathbf{a}_r, \mathbf{a}_a, \mathbf{a}_p) \;, \\
-    \forall (s \mapsto \mathbf{a}) \in \delta, (k \mapsto \mathbf{v}) \in \mathbf{a}_\mathbf{s}: &&C(s, \se_4(2^{32}-1) \frown k_{0\dots 27}) &\mapsto \mathbf{v} \;, \\
-    \forall (s \mapsto \mathbf{a}) \in \delta, (h \mapsto \mathbf{p}) \in \mathbf{a}_\mathbf{p}: &&C(s, \se_4(2^{32}-2) \frown h_{1\dots 28}) &\mapsto \mathbf{p} \;, \\
-    \forall (s \mapsto \mathbf{a}) \in \delta, (\tup{h, l} \mapsto \mathbf{t}) \in \mathbf{a}_\mathbf{l}: &&C(s, \se_4(l) \concat \hash(h)_{2\dots 29}) &\mapsto \se(\var{[\se_4(x) \mid x \orderedin \mathbf{t}]})
+    \forall (s \mapsto \mathbf{a}) \in \delta, (k \mapsto \mathbf{v}) \in \mathbf{a}_\mathbf{s}: &&C(s, \se_4(2^{32}-1) \frown k) &\mapsto \mathbf{v} \;, \\
```
zdave-parity: Service storage keys are hashed twice now; seems like it should be possible to avoid this? Maybe make service storage (`a_s`) bytes->bytes rather than hash->bytes?

text/accounts.tex
```diff
@@ -143,8 +143,8 @@ \subsection{Account Footprint and Threshold Balance}
 \begin{align}
   \forall a \in \mathcal{V}(\delta):\left\{\begin{aligned}
     a_i \in \N_{2^{32}} &\equiv 2\cdot|\,a_\mathbf{l}\,| + |\,a_\mathbf{s}\,| \\
-    a_o \in \N_{2^{64}} &\equiv \sum\limits_{\,(h, z) \in \keys{a_\mathbf{l}}\,} \!\!\!\!\!81 + z \\
-    &\phantom{\equiv\ } + \sum\limits_{x \in \mathcal{V}(a_\mathbf{s})} 32 + |x| \\
+    a_o \in \N_{2^{64}} &\equiv \sum\limits_{\,(h, z) \in \keys{a_\mathbf{l}}\,} \!\!\!\!81 + z \\
+    &\phantom{\equiv\ } + \sum\limits_{(x, y) \in a_\mathbf{s}} 34 + |y| + |x| \\
```
davxy, line 147: IIUC previously, it was $32$ bytes, corresponding to the old hashed key length. Now that’s been replaced by $|y|$. Where are the extra 34 bytes coming from?

text/merklization.tex
```diff
@@ -9,10 +9,10 @@ \subsection{Serialization}
 We define the state-key constructor functions $C$ as:
 \begin{equation}
   C\colon\left\{\begin{aligned}
-    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y_{27}} &\to \Y_{31} \\
+    \N_{2^8} \cup (\N_{2^8}, \N_S) \cup \tup{\N_S, \Y} &\to \Y_{31} \\
     i \in \N_{2^8} &\mapsto [i, 0, 0, \dots] \\
     (i, s \in \N_S) &\mapsto [i, n_0, 0, n_1, 0, n_2, 0, n_3, 0, 0, \dots]\ \where n = \se_4(s) \\
-    (s, h) &\mapsto [n_0, h_0, n_1, h_1, n_2, h_2, n_3, h_3, h_4, h_5, \dots, h_{26}]\ \where n = \se_4(s)
+    (s, h) &\mapsto [n_0, a_0, n_1, a_1, n_2, a_2, n_3, a_3, a_4, a_5, \dots, a_{26}]\ \where n = \se_4(s), a = \hash(h)_{0\dots 27}
```
davxy, line 15:
```suggestion
    (s, \mathbf{y}) &\mapsto [n_0, h_0, n_1, h_1, n_2, h_2, n_3, h_3, h_4, \dots, h_{26}]\ \where n = \se_4(s), h = \hash(\mathbf{y})_{0\dots 27}
```

- I'd rename $h$ to (bold) $\mathbf{y}$ (Just an arbitrary symbol)
- Then rename \$a\$ to \$h\$, as \$h\$ more intuitively suggests a "hash"

davxy:
@gavofyork  whops! I'm late :-/
FWIW I left some comments

text/accounts.tex
davxy: Perhaps this accounts for the worst-case (compact) encoded length prefix of both the key and value sequences? If that's the case, then
u64::MAX is encoded in 9 bytes, resulting in a total of 18 extra bytes. In that case, where do the additional 16 bytes come from?

PR #416, opened 2025-05-30
W* depends on ρ† and not ρ'

greywolve:
Could be wrong here, but it seems like we only need E_A and ρ† to get W.


gavofyork:
Will need resolving...

greywolve:
Resolved :)

PR #422, opened 2025-05-31
fix: digest code hash typo

prasad-kumkar:
Work digest and item code hash is defined as `h`, while being referred as `c` in serialization and 11.42

image

image

Changed to:
image

image


gavofyork:
Should be fixed now?

prasad-kumkar:
Yes, most have been resolved; thank you! There's still one instance in the digest code hash where it's referred to as x_c serialization

gavofyork:
That should also be fixed now

prasad-kumkar:
Looks good, thank you

PR #426, opened 2025-06-04
Pre-commit hook with `typos` step

Sajjon:
* Adjust a sentence in `text/auditing.tex`
* Add [pre-commit](https://pre-commit.com/) config file, with a single step running [`typos`](https://github.com/crate-ci/typos)
* Add `_typos.toml` a config file for `typos` to ignore some false positives, unused at the moment, since we have no false positives.
* Update README to mention setup installing pre-commit (N.B. `typos` need not be installed manually, `pre-commit` does that for us).


text/auditing.tex
```diff
@@ -97,7 +97,7 @@ \subsection{Selection of Reports}
...
 
-As $n$ increases with the passage of time $\mathbf{a}_n$ becomes known and defines our auditing responsibilities. We must attempt to reconstruct all work-packages and their requisite data corresponding to each work-report we must audit. This may be done through requesting erasure-coded chunks from one-third of the validators. It may also be short-cutted through asking a cooperative third-party (\eg an original guarantor) for the preimages.
+As $n$ increases with the passage of time $\mathbf{a}_n$ becomes known and defines our auditing responsibilities. We must attempt to reconstruct all work-packages and their requisite data corresponding to each work-report we must audit. This may be done through requesting erasure-coded chunks from one-third of the validators. It may also be shortcut by asking a cooperative third party (\eg an original guarantor) for the preimages.
```
Sajjon: "short-cutted" is non-standard, switched to use it as verb in past participle form.
"by asking" is more idiomatic than "through asking" I believe
"third-party" should not be hyphenated when used as a noun (only when used as a compound adjective before a noun, e.g., “a third-party service”).

Sajjon: Reverted `short-cutted` change.

PR #429, opened 2025-06-09
Various erasure coding fixes

zdave-parity:
- Typing errors.
- Out-of-place brackets.
- Argument to reconstruction function is a set and thus unordered.

Note that the unzip/lace functions have been removed as it's simpler to just use transpose with split/join.


davxy (approved):
LGTM

PR #437, opened 2025-06-18
Accumuluation: add missing c subscript in authqueue transition

daiagi:


text/accumulation.tex
```diff
@@ -208,7 +208,7 @@ \subsection{Execution}
       \forall c \in \coreindex :
         \local¬authqueue'\sub{c} &= (
           \accone(\mathbf{o}, \mathbf{w}, \mathbf{f}, \ps¬assigners\sub{c})_\ao¬poststate
-        )_\ps¬authqueue \\
+        )_\ps¬authqueue\sub{c} \\
```
gavofyork: Will need to be either `\subb{c}` or have additional parentheses.

gavofyork (changes requested):

text/accumulation.tex
daiagi: done

gavofyork (approved):

PR #457, opened 2025-07-01
Combine `accumulate` with `on_transfer`

### LINKED ISSUE #139: Consider: Combine `accumulate` with `on_transfer`
Thinking more about cross-service composition and better coordination over integration.

Queue `on_transfer` actions and allow accumulate to handle both WI outputs and transfers. This would mean transfer handlers could themselves send onward transfers and contribute to the accumulate (BEEFY) root.

Both of these may be important when we wish to acknowledge/enact integration outside of regular accumulation (since generally cross-service integration will rely on receiving a signalling transfer from the co-dependent service).

This throws up the uncomfortable question of gas management. But given our OrdAcc queuing system it might end up being ok.

gavofyork:
Closes #139

### DIFF
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -11,7 +11,7 @@
 
 Accumulation may be defined as some function whose arguments are $\justbecameavailable$ and $\accounts$ together with selected portions of (at times partially transitioned) state and which yields the posterior service state $\accountspostpreimage$ together with additional state elements $\stagingset'$, $\authqueue'$ and $\privileges'$.
 
-The proposition of accumulation is in fact quite simple: we merely wish to execute the \emph{Accumulate} logic of the service code of each of the services which has at least one work-digest, passing to it relevant data from said digests together with useful contextual information. However, there are three main complications. Firstly, we must define the execution environment of this logic and in particular the host functions available to it. Secondly, we must define the amount of gas to be allowed for each service's execution. Finally, we must determine the nature of transfers within Accumulate which, as we will see, leads to the need for a second entry-point, \emph{on-transfer}.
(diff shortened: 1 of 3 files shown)


text/serialization.tex
```diff
@@ -286,6 +286,11 @@ \subsection{Block Serialization}
     \var{\otX_\ot¬authtrace}
   }
   \\
+  \encode{\aiX \in \accinput} &\equiv \begin{cases}
```
zdave-parity, line 289: It's clear what you mean by this, but the formalisation seems a bit sketchy to me. If we have say an operand tuple `x`, it seems that both the `E(x in X)` and `E(x in I)` definitions would apply to `E(x)`.

text/accumulation.tex
```diff
@@ -158,77 +172,80 @@ \subsection{Execution}
...
+    &\tuple{\partialstate, \defxfers, \workreports, \dictionary{\serviceid}{\gas}} \to \tuple{\partialstate, \defxfers, \local¬fnservouts, \local¬fngasused} \\
+    &\tup{\psX, \mathbf{t}, \mathbf{r}, \mathbf{f}} \mapsto \tup{
       \tup{
         \local¬accounts', \local¬stagingset', \local¬authqueue', \local¬manager', \local¬assigners', \local¬delegator', \local¬alwaysaccers'
-      }, \concatall{\mathbf{t}}, \mathbf{b}, \mathbf{u}
+      }, \concatall{\mathbf{t}'}, \mathbf{b}, \mathbf{u}
     }\!\!\!\!\!\!\\
     &\text{where:}\\
     &\ \begin{aligned}
       \using \mathbf{s} &= \set{\build{
```
zdave-parity, line 189: `\mathbf{s}` needs to be a sequence as the order determines the order of the returned transfers. Guess this was also true before...

text/serialization.tex
gavofyork: Yes indeed; I don't believe either are encoded individually, so we can just have named functions for these.

text/accumulation.tex
gavofyork: It's a set. When used in sequence construction, it has an implicit order.

image

zdave-parity: That notation is not used here AFAICT. There is instead eg `s \orderedin \mathbf{s}` which I think is not defined when `\mathbf{s}` is a set. It could of course be defined, but I think it would be clearer to explicitly define an order for the services here?

PR #473, opened 2025-07-23
Misc: Small service IDs

### LINKED ISSUE #389: Consider: Small service IDs
Currently all services get a 32-bit pseudo-random ID which is hardly memorable.

Smaller IDs could be reserved and paid for. A simple way of doing this would be to allow service accounts to register a desired ID and introduce a new name-service privilege. The name-service privilege would allow for "renaming" any existing service to its desired service ID iff no account already exists at the desired service ID.

The gas costs for doing this would need to include the costs of moving over existing state keys. It would likely only be viable at service initialization.

A higher level "naming" service could look after determining exactly how shorter service IDs could be negotiated or purchased.

To allow for the smallest service indices to be used, we can move the Merkle tree encoding to the upper part of 32-bit space.

gavofyork:
Closes #389

### DIFF
--- a/text/accounts.tex
+++ b/text/accounts.tex
@@ -161,7 +161,7 @@ We may then define a third dependent term $\sa¬minbalance$, the minimum, or \em
 
 
 \subsection{Service Privileges}
-\Jam includes the ability to bestow privileges on a number of services. The portion of state in which this is held is denoted $\privileges$ and includes four kinds of privilege. The first, $\manager$, is the index of the \emph{manager} service which is the service able to effect an alteration of $\privileges$ from block to block as well as bestow services with storage deposit credits. The next, $\delegator$, is able to set $\stagingset$. Then $\registrar$ alone is able to create new service accounts with indices in the protected range. The following, $\assigners$, are the service indices capable of altering the authorizer queue $\authqueue$, one for each core.
+\Jam includes the ability to bestow privileges on a number of services. The portion of state in which this is held is denoted $\privileges$ and includes five kinds of privilege. The first, $\manager$, is the index of the \emph{manager} service which is the service able to effect an alteration of $\privileges$ from block to block as well as bestow services with storage deposit credits. The next, $\delegator$, is able to set $\stagingset$. Then $\registrar$ alone is able to create new service accounts with indices in the protected range. The following, $\assigners$, are the service indices capable of altering the authorizer queue $\authqueue$, one for each core.
 
 Finally, $\alwaysaccers$ is a small dictionary containing the indices of services which automatically accumulate in each block together with a basic amount of gas with which each accumulates. Formally:
 \begin{align}
--- a/text/accounts.tex
+++ b/text/accounts.tex
@@ -161,15 +161,21 @@ We may then define a third dependent term $\sa¬minbalance$, the minimum, or \em
 
 
 \subsection{Service Privileges}
-\Jam includes the ability to bestow privileges on a number of services. The portion of state in which this is held is denoted $\privileges$ and includes four kinds of privilege. The first, $\manager$, is the index of the \emph{manager} service which is the service able to effect an alteration of $\privileges$ from block to block as well as bestow services with storage deposit credits. The following, $\assigners$, are the service indices capable of altering the authorizer queue $\authqueue$, one for each core. The next, $\delegator$, is able to set $\stagingset$.
+\Jam includes the ability to bestow privileges on a number of services. The portion of state in which this is held is denoted $\privileges$ and includes four kinds of privilege. The first, $\manager$, is the index of the \emph{manager} service which is the service able to effect an alteration of $\privileges$ from block to block as well as bestow services with storage deposit credits. The next, $\delegator$, is able to set $\stagingset$. Then $\registrar$ alone is able to create new service accounts with indices in the protected range. The following, $\assigners$, are the service indices capable of altering the authorizer queue $\authqueue$, one for each core.
 
 Finally, $\alwaysaccers$ is a small dictionary containing the indices of services which automatically accumulate in each block together with a basic amount of gas with which each accumulates. Formally:
 \begin{align}
   \label{eq:privilegesspec}
-  \privileges \equiv \tuple{
-    \isa{\manager}{\serviceid} ,
-    \isa{\assigners}{\sequence[\Ccorecount]{\serviceid}} ,
-    \isa{\delegator}{\serviceid} ,
-    \isa{\alwaysaccers}{\dictionary{\serviceid}{\gas}}
-  }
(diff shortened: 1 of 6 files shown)


text/accounts.tex
```diff
@@ -161,15 +161,21 @@ \subsection{Account Footprint and Threshold Balance}
...
+\Jam includes the ability to bestow privileges on a number of services. The portion of state in which this is held is denoted $\privileges$ and includes four kinds of privilege. The first, $\manager$, is the index of the \emph{manager} service which is the service able to effect an alteration of $\privileges$ from block to block as well as bestow services with storage deposit credits. The next, $\delegator$, is able to set $\stagingset$. Then $\registrar$ alone is able to create new service accounts with indices in the protected range. The following, $\assigners$, are the service indices capable of altering the authorizer queue $\authqueue$, one for each core.
```
zdave-parity: s/four kinds/five kinds/

text/definitions.tex
```diff
@@ -269,6 +269,7 @@ \subsubsection{Constants}
   \item[$\Cslotseconds = 6$] The slot period, in seconds. See equation \ref{sec:epochsandslots}.
```
zdave-parity, line 270: Should add `\registrar` to the state components section above.

text/accumulation.tex
```diff
@@ -172,8 +173,10 @@ \subsection{Execution}
 \newcommand*{\local¬manager}{m}
 \newcommand*{\local¬assigners}{\mathbf{a}}
 \newcommand*{\local¬delegator}{v}
+\newcommand*{\local¬registrar}{r}
```
zdave-parity: These are now unused AFAICT?

text/definitions.tex
```diff
@@ -269,6 +269,7 @@ \subsubsection{Constants}
...
   \item[$\Cauthqueuesize = 80$] The number of items in the authorizations queue. See equation \ref{eq:authstatecomposition}.
   \item[$\Crotationperiod = 10$] The rotation period of validator-core assignments, in timeslots. See sections \ref{sec:coresandvalidators} and \ref{sec:workreportguarantees}.
+  \item[$\Cminpublicindex = 2^{16}$] The minimum public service index. Services of indices below these may only be created by the Registrar. See equation \ref{eq:newserviceindex}.
```
zdave-parity, line 273: There is no equation `newserviceindex` AFAICT.

text/merklization.tex
```diff
@@ -19,17 +19,17 @@ \subsection{Serialization}
 The state serialization is then defined as the dictionary built from the amalgamation of each of the components. Cryptographic hashing ensures that there will be no duplicate state-keys given that there are no duplicate inputs to $C$. Formally, we define $T$ which transforms some state $\thestate$ into its serialized form:
 \begin{equation}
   T(\thestate) \equiv \abracegroup{
-    &&C(1) &\mapsto \encode{\sq{\build{\var{x}}{x \orderedin \authpool}}} \;, \\
-    &&C(2) &\mapsto \encode{\authqueue} \;, \\
-    &&C(3) &\mapsto \encode{
+    &&C(2^{32} - 1) &\mapsto \encode{\sq{\build{\var{x}}{x \orderedin \authpool}}} \;, \\
```
zdave-parity: The first arg here can only be in 0..255. Not sure there's any need to change this?

text/pvm_invocations.tex
```diff
@@ -177,7 +177,7 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
       \is{\im¬yield}{\none},
       \is{\im¬provisions}{\sq{}}
     }\\
-    &\qquad\where \im¬nextfreeid = \text{check}((\decode[4]{\blake{\encode{\im¬id, \entropyaccumulator', \H_\¬timeslot}}} \bmod (2^{32}-2^9)) + 2^8) \\
+    &\qquad\where \im¬nextfreeid = \text{check}((\decode[4]{\blake{\encode{\im¬id, \entropyaccumulator', \H_\¬timeslot}}} \bmod (2^{32}-\Cminpublicindex-2^8)) + \Cminpublicindex) \\
```
zdave-parity, line 180: What's the `-2^8` for? I thought maybe something to do with trie key computation but I don't understand it if so.

text/pvm_invocations.tex
```diff
@@ -748,7 +750,7 @@ \subsection{Accumulate Functions}\label{sec:accumulatefunctions}
   \texttt{new} = 18 \\
   $g = 10$} &
   $\begin{aligned}
-    \using \sq{o, l, \sa¬minaccgas, \sa¬minmemogas, \sa¬gratis} &= \registers\subrange{7}{5} \\
+    \using \sq{o, l, \sa¬minaccgas, \sa¬minmemogas, \sa¬gratis, \local¬desiredid} &= \registers\subrange{7}{5} \\
```
zdave-parity: RHS needs adjusting to grab 6 regs.

text/pvm_invocations.tex
```diff
@@ -242,7 +242,7 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
 \begin{equation}
   \text{check}(i \in \serviceid) \equiv \begin{cases}
     i &\when i \not\in \keys{\local¬basestate_\ps¬accounts} \\
```
zdave-parity, line 245: Not changed by this PR but seems like `\local¬basestate` should really be passed as an argument here to avoid ambiguity.

text/merklization.tex
gavofyork: We opened up the bottom part of the service ID namespace; if we don't take action then there can be collisions.

text/pvm_invocations.tex
gavofyork: We reserve `Cminpublicindex` at the lower end of the namespace and 256 values at the top (to avoid collision between the inner account maps and the rest of the state state items in the trie).

text/merklization.tex
zdave-parity: I think the way things are defined there won't be collisions even if the full range of service IDs is used?

text/accumulation.tex
```diff
@@ -173,7 +173,6 @@ \subsection{Execution}
 \newcommand*{\local¬manager}{m}
 \newcommand*{\local¬assigners}{\mathbf{a}}
 \newcommand*{\local¬delegator}{v}
```
zdave-parity: All these other ones are unused too now I think?

text/pvm_invocations.tex
zdave-parity: Ok, I think this is not necessary but it doesn't really matter.

zdave-parity (approved):

PR #477, opened 2025-07-29
Order lastaccout before assign to new state

qiweiii:
add the missing ordering for last accumulation output

the last accumulation output type is sequence in 
image

and outer accumulate returns B which is a set

image
image

-------
Changes

- before:

image

- after:

image

### DIFF
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -360,7 +360,7 @@ Given the result of the top-level $\accseq$, we may define the posterior state $
   &\tup{
     \local¬numberofreportsaccumulated, \psX', \mathbf{b}, \local¬servicegasused
   } \equiv \accseq(g, \sq{}, \justbecameavailable^*, \psX, \alwaysaccers) \\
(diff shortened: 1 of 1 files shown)


text/accumulation.tex
```diff
@@ -358,8 +358,9 @@ \subsection{Final State Integration}
   \!\!\!\!\!\\
   \label{eq:finalstateaccumulation}
   &\tup{
-    \local¬numberofreportsaccumulated, \psX', \lastaccout', \local¬servicegasused
+    \local¬numberofreportsaccumulated, \psX', \mathbf{b}, \local¬servicegasused
   } \equiv \accseq(g, \sq{}, \justbecameavailable^*, \psX, \alwaysaccers) \\
+  &\lastaccout' \equiv \sqorderby{s}{\tup{s, h} \in \mathbf{b}} \\
```
gavofyork:
```suggestion
  &\lastaccout' \equiv \sq{\tup{s, h} \in \mathbf{b}} \\
```

PR #481, opened 2025-08-07
Serialization: Correct bit sequence encoding range

0xjunha:
In general, the Sigma sum notation (∑) uses inclusive start/end ranges. Other equations in GP follow this rule, e.g. [PVM instruction 102 & 103](https://graypaper.fluffylabs.dev/#/1c979cb/282102282102?v=0.7.1) and [Eq (H.9)](https://graypaper.fluffylabs.dev/#/1c979cb/3f71023f7102?v=0.7.1).

Updated bit sequence encoding function to follow this convention as well, ensuring it properly extracts 8 bits at once.

### DIFF
--- a/text/serialization.tex
+++ b/text/serialization.tex
@@ -67,7 +67,7 @@ A sequence of bits $b \in \bitstring$ is a special case since encoding each indi
   \encode{b \in \bitstring} &\equiv \begin{cases}
     \sq{} &\when b = \sq{} \\
     \sq{
-      \sum\limits_{i=0}^{\min(8, \len{b})}
+      \sum\limits_{i=0}^{\min(8, \len{b}) - 1}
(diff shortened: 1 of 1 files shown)


text/serialization.tex
```diff
@@ -67,7 +67,7 @@ \subsubsection{Bit Sequence Encoding}
   \encode{b \in \bitstring} &\equiv \begin{cases}
     \sq{} &\when b = \sq{} \\
     \sq{
-      \sum\limits_{i=0}^{\min(8, \len{b})}
+      \sum\limits_{i=0}^{\min(8, \len{b}) - 1}
```
gavofyork:
```suggestion
      \sum\limits_{i=0}^{i < \min(8, \len{b})}
```

PR #482, opened 2025-08-07
Adds explicit OOG check for each invocation mutator default case

ascrivener:


davxy:
Hmm... you're right. In our implementation, we don't return the invalid index error to the caller (aka `WHAT`).
Instead, we just treat the default case as a trap.   
We'll need to fix that in our implementation as well.

ascrivener:
spidey senses

gavofyork (approved):

davxy (approved):

PR #485, opened 2025-08-16
Remove page fault handling from host calls

ascrivener:
Host calls are careful not to do things that would cause a page fault, so this case never happens as the GP is currently written. Also removing related language.


text/pvm.tex
```diff
@@ -749,7 +745,7 @@ \subsection{Host Call Definition}
```
zdave-parity, line 745: Should also change the definition of `\contextmutator{X}` above to indicate that `\fault` is never returned.

text/pvm.tex
```diff
@@ -749,7 +745,7 @@ \subsection{Host Call Definition}
 
 However, when the exit reason to $\Psi$ is a host-call $\host$, then the resultant instruction-counter has a value of the host-call instruction and resuming with this state would immediately exit with the same result. Re-invoking would therefore require both the post-host-call machine state \emph{and} the instruction counter value for the instruction following the one which resulted in the host-call exit reason. This is always one greater plus the relevant argument skip distance. Resuming the machine with this instruction counter will continue beyond the host-call instruction.
 
-We use both values of instruction-counter for the definition of $\Psi_H$ since if the host-call results in a page fault we need to allow the outer environment to resolve the fault and re-try the host-call. Conversely, if we successfully transition state according to the host-call, then on resumption we wish to begin with the instruction directly following the host-call.
+If we successfully transition state according to the host-call, then on resumption we begin with the instruction directly following the host-call.
```
zdave-parity: I think more significant rewording of these paragraphs is needed. eg `On exit, the instruction counter $\imath'$ references the instruction \emph{which caused the exit}` does not seem accurate to me; AFAICT for host calls the instruction counter returned by `\Psi` points after the `ecalli` instruction. There is nothing in `\Psi_H` to step over the `ecalli` instruction AFAICT.

clearloop:
fyi [fuzz-reports/1755530535](https://github.com/davxy/jam-conformance/tree/main/fuzz-reports/0.6.7/traces/1755530535) triggers page fault at host call write, also, this is still possible to be triggered via hand made program API, e.g. skip jam-pvm-common, build programs with incorrect low-level api

text/pvm.tex
ascrivener: Actually, if the exit is halt or panic, \Psi returns 0 for the instruction counter. If the exit is OOG or page fault OR host-call, it returns the instruction counter for the instruction after the one that faulted.

And when the exit is a host-call, the host call instruction value is not stored in the instruction counter, but rather returned as a separate value with the exit reason.

I made a better version in the latest commit, lmk what you think

zdave-parity: For `\Psi_H` the instruction counter returned by `\Psi` only actually matters for the host-call case AFAICT. `\Psi_H` returns an instruction counter but this is never actually used? It should probably be removed. For that reason I probably wouldn't talk about any other case here.

The instruction counter returned by `\Psi` _does_ matter for the `invoke` case as it gets placed into the machine state. That stuff seems a little sketchy for at least two reasons:

- It appears that in the host-call case `invoke` explicitly steps over the `ecalli` instruction, but if `\Psi` is also doing this then it's being done twice which is surely wrong.
- If the returned instruction counter really is _after_ the failing instruction in the page-fault case then this is rather unhelpful, as obviously when we resume we want to retry the failing instruction rather than just skipping it. It's not clear to me from reading the PVM section what the intended behaviour is for page-faults. The text says "We define the memory-access exceptional execution state εµ which shall, if not ▸, singly effect the returned return of Ψ1 as following", but I don't really understand what is meant by "singly effect". It would be good to be more explicit here on what is returned in the other elements of the tuple (eg the instruction counter).

The out-of-gas case is a bit different to the page-fault case as it seems that we are supposed to actually execute the instruction which takes us over the gas limit. FWIW I don't think this matches PolkaJam behaviour, and is slightly surprising as it could lead to a significant gas overrun if for example the last instruction is an expensive host call.

gavofyork: In the case of the OOG, the instruction should not be executed, but the intent was to deduct the gas (largely out of safety since we don't distinguish between gas costs for actual instruction execution and gas calculation).

For page-faulting, the intention was to have $\Phi$ return an instruction counter value pointing to `ecalli`, rather than move beyond it prematurely since it is not known within $\Phi$ that the instruction will ever be "executed". $\Phi_H$ would be the one with a term describing the next instruction value beyond `ecalli` since it ensures that the host-call is processed successfully.

For `invoke`, we don't provide a means of the host-call handler (i.e. the `invoke` caller) from moving the instruction counter along by one since the instruction counter is opaque to that logic (we could provide another host-call for this reason but it seems a little extravagant). Instead we just move it along explicitly when we see the relevant execution status and make it a contract of the `invoke` function that the calling logic *must* handle the `ecalli` properly if it is to call `invoke` again on this context.

PR #488, opened 2025-09-02
PVM Invocations: Correct gas charge for `TRANSFER` host call failures

0xjunha:
Currently, `TRANSFER` host call charges additional gas `φ9` for all failure cases (`Panic`, `WHO`, `LOW`, `CASH`). 
However, for those cases, only the base gas `g = 10` should be charged since the additional gas is for covering the cost for processing deferred transfers. The failure cases do not add a deferred transfer to the accumulate context (**x**_**t**).

This PR corrects this by refactoring the logic to determine the host call result first, then the total gas charge. 
Two local variables are introduced: `c` for host call result code and `t` for transfer gas limit.

(before)
xfer_before

(after)
xfer_after

### DIFF
--- a/text/pvm_invocations.tex
+++ b/text/pvm_invocations.tex
@@ -813,7 +813,7 @@ Other than the gas-counter which is explicitly defined, elements of \textsc{pvm}
   \makecell*[l]{
   $\Omega_T(\gascounter, \registers, \memory, \imXY)$ \\
   \texttt{transfer} = 20 \\
-  $g = 10 + \registers_9 $} &
+  $g = 10 + t$} &
   $\begin{aligned}
     \using \sq{\dx¬dest, \dx¬amount, l, o} &= \registers\subrange{7}{4},  \\
     \using \mathbf{d} &= (\imX_\im¬state)_\ps¬accounts\\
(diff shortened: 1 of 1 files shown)


davxy:
Looks fine. Bit isn't simpler to leave everything as before and just add the variable t = l if (phi_7' = OK) else 0 ?

0xjunha:
Yes that will make equations simpler, but I wanted to avoid making `g` to depend on posterior state of the host call (`phi_7'`).
Screenshot 2025-09-02 at 2 18 19 PM

Here we first check OOG condition, then evaluate the host call results. If `g` depends on `phi_7'`, we cannot do OOG check before getting the final return value of the host function (including `phi_7'`), but `phi_7'` is not fully defined yet at that point.

davxy (approved):

davxy:
> Yes that will make equations simpler, but I wanted to avoid making `g` to depend on posterior state of the host call (`phi_7'`). Screenshot 2025-09-02 at 2 18 19 PM
> 
> Here we first check OOG condition, then evaluate the host call results. If `g` depends on `phi_7'`, we cannot do OOG check before getting the final return value of the host function (including `phi_7'`), but `phi_7'` is not fully defined yet at that point.

Makes sense

gavofyork:
$t$ is undefined in the case of a panic.

PR #491, opened 2025-09-10
Document Missing Dependencies in STF Equations

sierkov:
Adds missing STF dependencies to Section 4.2.1 for clarity.

Details:
- (4.16) Clarified that accumulation additionally depends on eta_prime (B.11).
- (4.14) Clarified that rho_prime additionally depends on:
  - beta_dagger (11.33)
  - alpha (11.29)
  - delta (11.30)
  - eta_prime (11.22)
- (4.11) Clarified that psi_prime additionally depends on kappa, lambda, and tau for signature verification (10.3), (10.5), (10.6).
- (4.5) Clarified that tau_prime depends on tau to enforce monotonicity (5.7).


davxy (dismissed):
> alpha (11.23)

Typo in the description. I imagine you refer to [11.29](https://graypaper.fluffylabs.dev/#/1c979cb/155602155602?v=0.7.1

sierkov:
@zdave-parity, @0xjunha If there’s consensus that dependencies required for signature verification can be omitted, I won’t insist on including them. This would remove kappa from (4.13), and kappa_prime, lambda_prime, and psi_prime from (4.14) in the proposed changes. I’ve updated the proposed changes and the description accordingly—please let me know if you have further feedback.

P.S. I’m traveling this week, so my responses may be delayed.

zdave-parity:
I think there is still a misunderstanding about what information the dependency graph is supposed to convey. The dependencies for some value x should include only the values required to actually compute x. That is, the values which (transitively) appear in the expression for x in the GP. Let's take `\thetime'` as an example, because it's so simple. That is defined like so:

    \thetime' \equiv \H_\¬timeslot

It requires only `\H_\¬timeslot`, which is part of `\theheader`, hence why only `\theheader` is listed as a dependency. Where did you get `\thetime` from?

PR #494, opened 2025-09-15
Keep full guarantees in availability assignments state (rho)

zdave-parity:
Full guarantees are needed for auditing:

- To determine the guarantors to try directly fetching the bundle from.
- In the case of a dispute, the guarantor signatures are needed to construct a disputes extrinsic.

It is convenient to have the full guarantees available in the state at the point where auditing begins. Without this, it would be necessary for validator nodes to track them separately from the point of reporting.

Note that `\mathbb{G}` is now used for the set of guarantees. Inner/guest PVM state uses `\mathbb{I}` instead.

Other minor fixes included in this PR:

- Fix order of set definitions. Intended to be alphabetical.
- Fix misnamed macros. \reportspostassurances and \reportspostguarantees were the wrong way around.
- Improve some of the wording around timing out of reports pending availability. Availability assignments are now proactively cleared when the timeout elapses.


text/auditing.tex
```diff
@@ -29,7 +29,7 @@ \subsection{Data Fetching}
...
+Each validator shall perform auditing duties on each valid block received. Since we are entering off-chain logic, and we cannot assume consensus, we henceforth consider ourselves a specific validator of index $v$ and assume ourselves focused on some recent block $\block$ with other terms corresponding to the state-transition implied by that block, so $\availassignments$ is said block's prior availability assignments, $\activeset$ is its prior validator set, $\header$ is its header \etc Practically, all considerations must be replicated for all blocks and multiple blocks' considerations may be underway simultaneously.
```
alxmirap, line 33:
```suggestion
Each validator shall perform auditing duties on each valid block received. Since we are entering off-chain logic, and we cannot assume consensus, we henceforth consider ourselves a specific validator of index $v$ and assume ourselves focused on some recent block $\block$ with other terms corresponding to the state-transition implied by that block, so $\availassignments$ is said block's prior availability assignments, $\activeset$ is its prior validator set, $\header$ is its header \etc. Practically, all considerations must be replicated for all blocks and multiple blocks' considerations may be underway simultaneously.
```

alxmirap: Perhaps this could be rephrased to be a little less ambivalent. \availassignments is called the "pending work-report guarantees" in the glossary, and calling them here "availability assignments" may lead to confusion of whether these are two distinct concepts. 
In section 11.1, this is further defined as "the work-report guarantees which have been re-
ported but are not yet known to be available to a super-
majority of validators". The term "availability assignment" is so far only used three times in the text, and there is not really any one that defines it. Perhaps we should name it so in section 11.1.

text/auditing.tex
```diff
@@ -176,7 +176,7 @@ \subsection{Selection of Reports}
...
 
-As a block author, we include a judgment extrinsic which collects judgment signatures together and reports them on-chain. In the case of a non-valid judgment (\ie one which is not two-thirds-plus-one of judgments confirming validity) then this extrinsic will be introduced in a block in which accumulation of the non-valid work-report is about to take place. The non-valid judgment extrinsic removes it from the pending work-reports, $\reports$. Refer to section \ref{sec:disputes} for more details on this.
+As a block author, we include a disputes extrinsic which collects judgment signatures together into verdicts and reports them on-chain. In the case of a non-valid verdict (\ie one which is not two-thirds-plus-one of judgments confirming validity) then this verdict will be introduced in a block in which accumulation of the non-valid work-report is about to take place. The non-valid verdict in the disputes extrinsic clears its availability assignment from $\availassignments$. Refer to section \ref{sec:disputes} for more details on this.
```
alxmirap: I suppose "non-valid" here is not used to say "a verdict that is not valid" (which would probably be 'invalid verdict') but rather a "verdict that establishes a report is not valid". In that case, I believe the parentheses are too broad, as they allow any mal-formed verdict (eg, one with fewer than 2/3V+1 votes) and should instead be something like "one which is two-thirds-plus-one judgements refuting validity".

text/reporting_assurance.tex
```diff
@@ -12,18 +12,20 @@ \section{Reporting and Assurance}
 
 
 \subsection{State}
-The state of the reporting and availability portion of the protocol is largely contained within $\reports$, which tracks the work-reports which have been reported but are not yet known to be available to a super-majority of validators, together with the time at which each was reported. As mentioned earlier, only one report may be assigned to a core at any given time. Formally:
+The state of the reporting and availability portion of the protocol is largely contained within $\availassignments$, which tracks the work-report guarantees which have been reported but are not yet known to be available to a super-majority of validators, together with the time at which each was reported. As mentioned earlier, only one guarantee may be assigned to a core at any given time. Formally:
```
alxmirap: This is where I think we should name the term "availability assignment" somehow. However, it is not clear to me why this wording was chosen. Is it because we assign each guarantee to a core, while it is unavailable? And iiuc an entry stops being in \availassignments once it becomes available?

text/auditing.tex
davxy: These are the "work report guarantees with pending availability"- a bit of a mouthful.

I agree that we should use a consistent term throughout the spec. We're so used to this terminology - *pending availability assignments*, *pending reports* (now guarantees) - that it may not feel confusing to us, but it could be for a first-time reader.  
Perhaps *pending guarantees* is fine to be used throughout the spec

text/reporting_assurance.tex
davxy: The state item ($\rho$) holds report-guarantees (one per core) while waiting for the corresponding report to become available.  
When the report becomes available (i.e., upon receiving the corresponding availability extrinsic), we remove the guarantee from the bucket, freeing space for another guarantee that may arrive via the guarantees extrinsic, possibly within the same block.  
We process the availability extrinsic first, before handling guarantees in that block.

text/auditing.tex
davxy:
```suggestion
As a block author, we include a disputes extrinsic which collects judgment signatures into verdicts and reports them on-chain. If a verdict establishes that a work report is not valid (\ie, fewer than two-thirds-plus-one of judgments confirming report validity) then it is included in a block where accumulation of the corresponding work-report is about to take place. This verdict also clears the corresponding pending guarantee from $\availassignments$. Refer to section \ref{sec:disputes} for more details on this.
```

Ref https://github.com/gavofyork/graypaper/pull/494/files#r2362521608

davxy (dismissed):
Looks good

text/auditing.tex
zdave-parity: I agree we should be consistent with terminology. I think "availability assignments" is a good term, as these are the guarantees which are currently assigned to each core to be made available. I took this term from PolkaJam, and according to `git blame` Gav named the `AvailabilityAssignment` struct, so it has some sort of blessing :P. I'll see if I can use it more consistently...

text/auditing.tex
zdave-parity: > I suppose "non-valid" here is not used to say "a verdict that is not valid" (which would probably be 'invalid verdict') but rather a "verdict that establishes a report is not valid".

Yes.

> In that case, I believe the parentheses are too broad, as they allow any mal-formed verdict

The parentheses seem fine to me: any verdict which is not a good verdict should clear the work-report from the availability assignments. Malformed verdicts, that mean the block is not valid, don't really matter here because if the block is invalid then its posterior state is irrelevant. But wonky verdicts do matter, and they wouldn't fit your description. They should cause availability assignments to be cleared in the same way that bad verdicts do.

text/auditing.tex
zdave-parity: Hopefully better now.

text/reporting_assurance.tex
zdave-parity: This now uses the term "availability assignment".

text/auditing.tex
zdave-parity: Reworded this. Hopefully clearer now.

alxmirap (commented):
LGTM

gavofyork (changes requested):
Needs rebasing...

zdave-parity:
Rebased.

Also got rid of `\accinput`. I hadn't realised this was also using `\mathbb{I}` as it wasn't listed in the notation section. It's only really needed in one place (definition of the accumulation invocation function); it's simpler to just write `U cup X` there instead.

This will likely need rebasing again on top of #514, or the other way around. I'll do that on a separate branch for now...

zdave-parity:
> This will likely need rebasing again on top of #514, or the other way around. I'll do that on a separate branch for now...

Branch merging this with #514 available [here](/gavofyork/graypaper/tree/dave/guarantees-in-state-plus-var-vals).

gavofyork:
@zdave-parity what's the preferred way forward here? Merge #514 first and then rebase this to `main` and review? Or merge the new branch into #514 and then review/merge them both in one go?

zdave-parity:
> @zdave-parity what's the preferred way forward here? Merge #514 first and then rebase this to `main` and review? Or merge the new branch into #514 and then review/merge them both in one go?

Probably best to merge one at a time. Either way round works, I've done the merge already so should be trivial to update either PR after the other goes in.

zdave-parity:
Just merged main branch. Think this is ok to submit now.

gavofyork (approved):

PR #496, opened 2025-09-17
Safrole refinement

davxy:
- Refer to $\iota$, $\kappa$, $\lambda$ $\gamma_P$ as sequences
  - in `preamble.tex` these are still aliased to `pendingset`, `stagingset` etc, but I'm not sure if is worth renaming to `pendingseq` or `pendingvals` or whatever.
- Header contains two VRF signatures (seal and entropy source)
- Some minor nits in both Safrole and cryptography section


text/notation.tex
```diff
@@ -158,15 +158,15 @@ \subsubsection{Hashing}
...
+$\edsignature{k}{m \in \blob} \subset \blob[64]$ is the set of valid Ed25519 signatures, defined by \cite{rfc8032}, made through knowledge of a secret key whose public key counterpart is $k \in \hash$ and whose message is $m$. To aid readability, we denote the set of valid public keys $\edkey$.
 
-We denote the set of valid Bandersnatch public keys as $\bskey$, defined in appendix \ref{sec:bandersnatch}. $\bssignature{k \in \bskey}{x \in \blob}{m \in \blob} \subset \blob[96]$ is the set of valid singly-contextualized signatures of utilizing the secret counterpart to the public key $k$, some context $x$ and message $m$.
+We denote the set of valid Bandersnatch public keys as $\bskey$, defined in appendix \ref{sec:bandersnatch}. $\bssignature{k \in \bskey}{x \in \blob}{m \in \blob} \subset \blob[96]$ is the set of valid singly-contextualized \textsc{vrf} signatures of utilizing the secret counterpart to the public key $k$, some context $x$ and message $m$.
```
zdave-parity: of utilizing -> utilizing?

text/notation.tex
```diff
@@ -158,15 +158,15 @@ \subsubsection{Hashing}
...
+$\bsringproof{r \in \ringroot}{x \in \blob}{m \in \blob} \subset \blob[784]$, meanwhile, is the set of valid Bandersnatch Ring\textsc{vrf} deterministic singly-contextualized proofs of knowledge of a secret within some set of secrets identified by some root in the set of valid \emph{roots} $\ringroot \subset \blob[144]$. We denote $\getringroot{\mathbf{s} \in \sequence{\bskey}} \in \ringroot$ to be the root specific to the set of public key counterparts $\mathbf{s}$. A root implies a specific sequence of Bandersnatch key pairs, knowledge of one of the secrets would imply being capable of making a unique, valid---and anonymous---proof of knowledge of a unique secret within the set.
```
zdave-parity: Will take your word for it that the root implies a particular order. "Set" is used a few times here, seems like more of these should be changed to sequence?

text/safrole.tex
```diff
@@ -5,13 +5,13 @@ \section{Block Production and Chain Growth}
...
+In order to generate this sequence of tickets in regular operation, and in particular to do so without making public the correspondence relation between them and the validator key sequence, we use a novel Ring\textsc{vrf} cryptographic scheme, built on the Bandersnatch curve. This scheme allows for a proof to be provided which simultaneously guarantees the author controlled a key within a sequence (in our case validators), and secondly provides an output, an unbiasable deterministic hash giving us a secure verifiable random function (\textsc{vrf}). This anonymous and secure random output is a \emph{ticket} and validators' tickets with the best score define the sealing tickets allowing the associated validators to exercise their privilege and create a new block at the appropriate timeslot.
```
zdave-parity: The wording of the last sentence here is not particularly clear. Would be useful to name the ticket sequence in the paragraph above where it is introduced, to allow it to be clearly referred to here.

text/safrole.tex
```diff
@@ -78,7 +78,7 @@ \subsection{Safrole Basic State}
   }
 \end{align}
 
-As we state in section \ref{sec:sealandentropy}, Safrole requires that every block header $\H$ contain a valid seal $\H_\¬sealsig$, which is a Bandersnatch signature for a public key at the appropriate index $m$ of the current epoch's seal-key series, present in state as $\sealtickets$.
+As we state in section \ref{sec:sealandentropy}, Safrole requires that every block header $\H$ contain a valid seal $\H_\¬sealsig$, which is a Bandersnatch signature produced with the private key corresponding to the entry at index $m$ of the current epoch's seal-key series, present in state as $\sealtickets$.
```
zdave-parity: s/seal-key/ticket/

text/safrole.tex
```diff
@@ -139,7 +139,7 @@ \subsection{Key Rotation}
 \subsection{Sealing and Entropy Accumulation}
 \label{sec:sealandentropy}
 
-The header must contain a valid seal and valid \textsc{vrf} output. These are two signatures both using the current slot's seal key; the message data of the former is the header's serialization omitting the seal component $\H_\¬sealsig$, whereas the latter is used as a bias-resistant entropy source and thus its message must already have been fixed: we use the entropy stemming from the \textsc{vrf} of the seal signature. Formally:
+The header must contain a valid seal and valid entropy source. These are two \textsc{vrf} signatures both using the current slot's seal key; the message data of the former is the header's serialization omitting the seal component $\H_\¬sealsig$, whereas the latter is used as a bias-resistant entropy source and thus its message must already have been fixed: we use the entropy stemming from the \textsc{vrf} of the seal signature. Formally:
```
zdave-parity: Is "seal key" defined anywhere? Would maybe say something like "using the private key corresponding to the current slot's ticket" instead, though this is obviously a bit wordier.

zdave-parity (commented):
Looks reasonable to me. I would be in favour of renaming eg `\activeset` to `\activevals`. I think this would be clearer and also avoid implying that they are sets rather than sequences.

text/notation.tex
davxy: Yeah, order matters. I can change to sequence

text/safrole.tex
davxy: I renamed it to *slot-sealer series*. This matches the terminology used just above for  $\gamma_S$.
It represents a sequence of public keys or a sequence of tickets, each of which implicitly and anonymously maps to a public key.

text/safrole.tex
davxy: I have reworked the paragraph a bit. The terms "seal-key" sequence and "slot-sealer" sequence were used somewhat interchangeably, but I decided to use the latter consistently. The reason is that the word "key" might suggest we are referring to an actual key, whereas an entry in the "slot-sealer" sequence can represent either a key or a ticket (which implicitly maps to a key). I have made this clarification explicitly elsewhere.

text/safrole.tex
```diff
@@ -78,7 +78,7 @@ \subsection{Safrole Basic State}
   }
 \end{align}
 
-As we state in section \ref{sec:sealandentropy}, Safrole requires that every block header $\H$ contain a valid seal $\H_\¬sealsig$, which is a Bandersnatch signature for a public key at the appropriate index $m$ of the current epoch's seal-key series, present in state as $\sealtickets$.
+As we state in section \ref{sec:sealandentropy}, Safrole requires that every block header $\H$ contain a valid seal $\H_\¬sealsig$, which is a Bandersnatch signature produced with the private key corresponding to the entry at index $m$ of the current epoch's slot-sealer sequence $\sealtickets$.
```
0xjunha: Should we use `m'` and `γS'` here? Or fine to keep as-is since this is a general explanation

text/safrole.tex
0xjunha: Bandersnatch key root -> ring root would be more consistent

0xjunha: from the validator key `set` -> `sequence` for consistency

text/safrole.tex
```diff
@@ -280,11 +280,11 @@ \subsection{The Markers}
...
+The extrinsic $\xttickets$ is a sequence of proofs of valid tickets; a ticket implies an entry in our epochal ``contest'' to determine which validators are privileged to author a block for each timeslot in the following epoch. Tickets specify an entry index (a natural number less than $\Cticketentries$) together with a proof of ticket's validity. The proof implies a ticket identifier, a high-entropy unbiasable 32-octet sequence, which is used both as a score in the aforementioned contest and as input for the block's entropy source $\textsc{vrf}$ signature.
 
 Towards the end of the epoch (\ie $\Cepochtailstart$ slots from the start) this contest is closed implying successive blocks within the same epoch must have an empty tickets extrinsic. At this point, the following epoch's seal key sequence becomes fixed.
```
0xjunha: seal key -> slot-sealer

text/definitions.tex
0xjunha: sealing-key -> slot-sealer

0xjunha (commented):
Added some comments

text/notation.tex
```diff
@@ -158,15 +158,15 @@ \subsubsection{Hashing}
...
+$\edsignature{k}{m \in \blob} \subset \blob[64]$ is the set of valid Ed25519 signatures, defined by \cite{rfc8032}, made through knowledge of a secret key whose public key counterpart is $k \in \hash$ and whose message is $m$. To aid readability, we denote the set of valid public keys $\edkey$.
 
-We denote the set of valid Bandersnatch public keys as $\bskey$, defined in appendix \ref{sec:bandersnatch}. $\bssignature{k \in \bskey}{x \in \blob}{m \in \blob} \subset \blob[96]$ is the set of valid singly-contextualized signatures of utilizing the secret counterpart to the public key $k$, some context $x$ and message $m$.
+We denote the set of valid Bandersnatch public keys as $\bskey$, defined in appendix \ref{sec:bandersnatch}. $\bssignature{k \in \bskey}{x \in \blob}{m \in \blob} \subset \blob[96]$ is the set of valid singly-contextualized \textsc{vrf} signatures utilizing the secret counterpart to the public key $k$, some context $x$ and message $m$.
```
0xjunha, line 167: Here, context of Bandersnatch signatures/proofs are of blob type, whereas in Appendix G. They are hash types. Also would be nice to use `c` instead of `x`?

text/notation.tex
0xjunha, line 173: It seems the exact type (blob length) of BLS signatures is not specified anywhere. Would be nice to add in Appendix I.

0xjunha:
Added a commit to fix VRF outputs from `Y(..) ⊂ H` to `Y(..) ∈ H`

text/safrole.tex
gavofyork: Yes this rewording needs rethinking. It introduces the term *ticket* ("This anonymous and secure random output is a ticket") after its first use ("In order to generate this sequence of tickets in regular operation").

gavofyork (changes requested):
Needs rebasing...

Also at least one conversation point to be resolved (definition of tickets).

text/safrole.tex
davxy: @zdave-parity I reworked a bit the paragraph https://github.com/gavofyork/graypaper/pull/496/changes/905f0f8a40bc1b1bc671b21fe69621918497926e

text/safrole.tex
davxy: https://github.com/gavofyork/graypaper/pull/496/changes/b1821347d5066dbd31f01543a52410d5a830d61c

text/safrole.tex
davxy: https://github.com/gavofyork/graypaper/pull/496/changes/f856ac21331156193d15024607fea3b0dd8cbee4

text/safrole.tex
davxy: https://github.com/gavofyork/graypaper/pull/496/changes/e52310823df57e56209bbc5858eaad3f5d1d0cbb

text/safrole.tex
davxy: https://github.com/gavofyork/graypaper/pull/496/changes/e52310823df57e56209bbc5858eaad3f5d1d0cbb

text/definitions.tex
davxy: https://github.com/gavofyork/graypaper/pull/496/changes/e52310823df57e56209bbc5858eaad3f5d1d0cbb

text/notation.tex
davxy: I imagine because is not yet fully defined. I don't think this will be a normal BLS signature for the same reason this is not a normal bls key (if we're it is a BLS key ++ ECDSA key). I leave this as-is as not related to safrole

gavofyork (approved):

PR #497, opened 2025-09-17
Account gas for memory access

davxy:
Charging for the instruction that triggered the trap makes sense also according to this statement at the begin of section A.1

> Assuming the program blob is valid (which can be validated statically), some gas is always charged whenever execution
> is attempted. This is the case even if no instruction is effectively executed and machine state is unchanged (i.e. the result
> state is equal to the parameter).

Otherwise no gas is charged if the very first instruction is one that triggers a trap for memory access.

---

Furthermore this is consistent and equivalent to an explicit `trap` instruction, which is charged 1 gas unit.


davxy:
After

Screenshot from 2025-09-17 16-46-57

0xjunha (approved):

Issue #498, opened 2025-09-23
The calculation of g: The number of reports guaranteed by the validator


YCC3741:
When testing [trace/preimage_light/0000008](https://github.com/davxy/jam-test-vectors/blob/master/traces/preimages_light/00000008.json), we found issues with [pi_V[v]_g](https://graypaper.fluffylabs.dev/#/38c4e62/19ea0119fc01?v=0.7.0) as follows.

In tiny mode, the parameters are `R = 4` and `E = 12`, which means the cores where the guarantors for slot 7 and slot 8 reside will rotate.

**What we observed**

- Initially, our `pi_V[v]_g` was 2.  
- In slot 8, we received a `guarantee_extrinsic` containing two elements belonging to slot 7 and slot 8, respectively.  
- Core assignments:
  - slot 8 — `[0,1,1,0,0,1]`
  - slot 7 — `[1,0,0,1,1,0]`
- Based on these assignments, we confirmed that the reports received for both slots were valid.

**What we did**

- We incremented `pi_V[v]_g` twice by +1 (since `kappa_v_prime ∈ G` per GP 0.7.0 [formula 13.5](https://graypaper.fluffylabs.dev/#/38c4e62/19f40119f401?v=0.7.0)), resulting in `pi_V[v]_g_prime = 4`.

**Where the ambiguity is**

- GP 0.7.0 [formula 13.5](https://graypaper.fluffylabs.dev/#/38c4e62/19f40119f401?v=0.7.0) appears to only check whether `kappa_v_prime ∈ G`.
- However, this seems to conflict with the description above 13.3:
  - [**g:** The number of reports guaranteed by the validator.](https://graypaper.fluffylabs.dev/#/38c4e62/197700197a00?v=0.7.0)

Should both reports be counted toward `g` in this case, or should we follow formula 13.5 strictly and add +1 only once?

**Proposal**

This ambiguity likely requires us to rethink how we calculate `g`.

PR #499, opened 2025-09-23
chore: the calculation of g: The number of reports guaranteed by the …

YCC3741:
As the issue #498 
Perhaps we should revisit the defination of the calculation of g: The number of reports guaranteed by the validator

### DIFF
--- a/text/statistics.tex
+++ b/text/statistics.tex
@@ -60,7 +60,7 @@ The objective statistics are updated in line with their description, formally:
         0 &\otherwise
       \end{cases}\\
     \valstatsaccumulator'\subb{v}_\vs¬guarantees &\equiv
-      \mathbf{a}\subb{v}_\vs¬guarantees + \sum_{\activeset'\sub{v} \in \reporters}{1}\\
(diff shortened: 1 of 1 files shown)


text/statistics.tex
```diff
@@ -60,7 +60,7 @@ \subsection{Validator Activity}
         0 &\otherwise
       \end{cases}\\
     \valstatsaccumulator'\subb{v}_\vs¬guarantees &\equiv
-      \mathbf{a}\subb{v}_\vs¬guarantees + (\activeset'\sub{v} \in \reporters)\\
+      \mathbf{a}\subb{v}_\vs¬guarantees + \sum_{\activeset'\sub{v} \in \reporters}{1}\\
```
zdave-parity: This is totally bogus

YCC3741: But if you receive a guarantee_extrinsic with two elements, each belonging to a different slot, doesn't it count as two?

YCC3741: Oh I see. I will rewrite it later. Thx

PR #500, opened 2025-09-29
Account for gas reserved by transfer and always acc items

davxy:
- When computing $i$, deduct the gas already reserved for *Transfers* and *Always Acc* jobs.  
  - In the current version, we might schedule more reports than can actually be processed, since *Transfers* and *Always Acc* are executed first.  
  - This adjustment is consistent, as we add the gas limits for *Transfers* and *Always Acc* to the gas usable by the $\Delta^+$ call.  
- Use the gas limit for *Transfers* obtained from the output of $\Delta_*$, to the recursive $\Delta_+$ call.

### Before

Screenshot from 2026-05-25 19-05-53

### After

Screenshot from 2026-05-25 19-15-27


zdave-parity (approved):

text/accumulation.tex
```diff
@@ -160,11 +160,12 @@ \subsection{Execution}
       \tup{i + j, \psX', \mathbf{b}^* \!\cup \mathbf{b}, \mathbf{u}^* \!\!\concat \mathbf{u}}\!\!\!\! &
         \text{o/w}\!\!\!\!\!\!\!\! \\
     \end{cases} \\
-    &\quad\where i = \max(\Nmax{\len{\mathbf{r}} + 1}): \sum_{r \in \mathbf{r}\sub{\dots i}, d \in r_\wr¬digests}(d_\wd¬gaslimit) \le g \\
-    &\quad\also n = \len{\mathbf{t}} + i + \len{\mathbf{f}} \\
```
gavofyork, line 165: Why?

davxy: The gas budget $g$ given to $\Delta_+$ is shared between all three things accumulated in a round:
- $i$ work reports
- deferred transfers **t** (zero in the top-level call)
- always-acc services **f** (non-zero only in top-level call)

Inside the $\Delta_*$ transfers and always-acc items consume gas first, so the budget actually available for the $i$ work reports is $g$ minus what those two consume.

With the old condition, only the work-reports' gas limits had to fit in $g$, so we could pick an $i$ for which there isn't enough gas left after transfers and always-acc are paid for. So the reports selected here might not actually be all processable.

It is also consistent with the recursive call: $g^* = g + \sum_{t \in \mathbf{t}^*}(t_g) - \sum (u)$ adds back the gas budgeted for the transfers to be processed in the next round, which makes sense if this is taken into consideration when computing $i$ in the inner round.

Note that when we compute the gas for the next round g^* we don't account for the always-acc gas as these are already processed in the top-level $\Delta_+$

davxy: FWIW This logic is already what polkajam implements

gavofyork: Why do we add the transfers budget to `g^*` only to offset it again against parameter $g$ the invocation of $\Delta^+$? Surely it's simpler to consider all transfers free and avoid introducing their gas budget at all?

davxy: For the choice of $i$ with a single round yes: the $+Σ t_g$ added into $g^*$ is exactly the $Σ t_g$ subtracted in the next round's selection inequality, so there it cancels.

But $g^*$ is the **running budget threaded through every later round**, and the two are not symmetric:

- the $+Σ t_g$ enters g* and **stays there for all subsequent rounds**;
- the $−Σ t_g$ appears only inside the next round's selection inequality and never decrements $g^*$

So it cancels for that one round's $i$, not for the carried budget. It has to be added back because $\Delta_*$ processes the incoming transfers unconditionally (g only gates how many work reports $i$ we can take), and each deferred transfer carries sender-funded gas. Carrying that gas forward is what reserves it for the round that actually executes the transfer.

---

A 3-round example makes the difference visible.

- N: formalism introduced by this PR
- D: formalism with dropped transfer terms

**Round 1**: Initial budget g = 100, no always-acc, no incoming transfers. Selected reports use 60 and emit transfers reserving 30 for round 2.
- N: $g^* = 100 - 60 + 30 = 70$
- D: $g^* = 100 - 60 = 40$

**Round 2**: Process the 30 of transfers. Bound for reports:
- N: $wr ≤ 70 − 30 = 40$
- D: $wr ≤ 40$

Transfers consume 20, reports consume 20, so used = 40, and they emit transfers reserving 10 for round 3.
- N: $g^* = 70 - 40 + 10 = 40$
- D: $g^* = 40 - 40 = 0$   

**Round 3**: Process the 10 of transfers. Bound for reports: 
- N: $wr ≤ 40 − 10 = 30$ (still room for work-reports)
- D $wr ≤ 0$ (only transfers are processed)

PR #502, opened 2025-10-07
Add back processed transfer count to service statistics

zdave-parity:
This was removed when accumulate was combined with on_transfer. It seems like useful statistic, so add it back.

### DIFF
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -5,6 +5,7 @@
 \newcommand*{\local¬fnprovide}{I}
 \newcommand*{\local¬numberofreportsaccumulated}{n}
 \newcommand*{\local¬servicegasused}{\mathbf{u}}
+\newcommand*{\local¬processedtransfers}{\mathbf{t}}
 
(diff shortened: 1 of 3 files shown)


davxy:
In 13.7, $a$ domain need to be fixed:

Currently:

Screenshot from 2025-10-08 12-13-27

But according to **S** (12.29) needs to be $a \in (\mathbb{N}_G, \mathbb{N}, \mathbb{N})$

zdave-parity:
It's somewhat disappointing that no one has pointed this error out. It suggests that people are simply updating their implementations to match PolkaJam rather than the GP.

jaymansfield:
> It suggests that people are simply updating their implementations to match PolkaJam rather than the GP.

Not really at all. If an implementation encoded C(13) according to the πS definition in 13.7 (with gas last), they are following what is defined in the GP, not PolkaJam.

davxy (approved):

gavofyork (approved):

PR #510, opened 2026-01-15
Clarify missing validator-key subscript in auditing and beefy equations

yu2C:
### current main
image

image

### after fix
image

image

In equations (17.17) and (18.1), the validator index `v` was previously
encoded implicitly in the subscript of the signing operator `\fnsigndata`. Under the
current LaTeX setup, this caused the index to be dropped or rendered
ambiguously.

Note: the `\bar{}` and `BLS` notation shift due to the definitions in `preamble.tex` [line350](https://github.com/gavofyork/graypaper/blob/d3512eddbb55bc294fad43591808848c256215cb/preamble.tex#L350) and [line194](https://github.com/gavofyork/graypaper/blob/d3512eddbb55bc294fad43591808848c256215cb/preamble.tex#L194)

### DIFF
--- a/preamble.tex
+++ b/preamble.tex
@@ -191,7 +191,7 @@
 \newcommand*{\fnzeropad}[1]{\mathcal{P}_{#1}}
 \newcommand*{\fnseqfromhash}[1]{\mathcal{Q}_{#1}}
 \newcommand*{\fnecrecover}{\mathcal{R}}
-\newcommand*{\fnsigndata}[1]{\mathcal{S}_{#1}}
+\newcommand*{\fnsigndata}{\mathcal{S}}
 \newcommand*{\fnsubifnone}{\mathcal{U}}
 \newcommand*{\fnsext}[1]{\mathcal{X}_{#1}}
 \newcommand*{\fnbanderout}{\mathcal{Y}}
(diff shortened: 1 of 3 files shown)


zdave-parity:
I agree the current rendering is broken. This appears to be a general problem with `\edsigndata` and `\blssigndata`; the signing key is sometimes not rendered properly. I'm not sure why but pulling the signing key out of the `\bar`/`\accentset` seems to solve the problem, ie changing the definitions in `preamble.tex` as follows:

    -\newcommand*{\fnsigndata}[1]{\mathcal{S}_{#1}}
    +\newcommand*{\fnsigndata}{\mathcal{S}}

    -\newcommand*{\edsigndata}[2]{\bar{\fnsigndata{#1}}\left(#2\right)}
    -\newcommand*{\bssigndata}[2]{\accentset{\backsim}{\fnsigndata{#1}}\left(#2\right)}
    -\newcommand*{\blssigndata}[2]{\accentset{\mathrm{B\!L\!S}}{\fnsigndata{#1}}\left(#2\right)}
    +\newcommand*{\edsigndata}[2]{\bar{\fnsigndata}_{#1}\left(#2\right)}
    +\newcommand*{\blssigndata}[2]{\accentset{\mathrm{B\!L\!S}}{\fnsigndata}_{#1}\left(#2\right)}

This also fixes the weird placement of the bar/BLS.

FWIW I agree with the addition of the previously missing `\vk¬bls`.

zdave-parity (approved):
Thanks, this looks reasonable to me.

davxy (approved):

PR #514, opened 2026-02-18
Support smaller validator sets

zdave-parity:
The designate host call now has a length argument, permitting validator sets smaller than 1023 validators to be specified. For simplicity, we require that the size is a multiple of 3 and at least 6 (i.e. 2 cores).

The number of cores remains fixed at 341, but the number of "active" cores scales with the number of active validators. It is not possible to report work-packages on inactive cores.

When there are fewer validators, we permit each validator to generate more Safrole tickets, to ensure the accumulator can be filled.

For simplicity, the erasure coding rate and piece size are defined such that segments are always a whole number of pieces and do not require padding. In practice this means that the erasure coding rate is suboptimal for many validator set sizes; it is recommended that the validator set size is picked from the set of optimal sizes given in the erasure coding section.


zdave-parity:
Just rebased.

text/erasure_coding.tex
```diff
@@ -4,17 +4,24 @@ \section{Erasure Coding}
...
+The foundation of the data-availability and distribution system of \Jam is a systematic Reed-Solomon erasure coding function in \textsc{gf}($2^{16}$), the same transform as done by the algorithm of \cite{lin2014novel}. The coding rate is $\fnecoriginalshards(v)$:$v$, where $v \in \valcount$ is the number of validators.
+\begin{equation}\label{eq:ecoriginalshards}
+  \fnecoriginalshards(v \in \valcount) \equiv \max\left(\set{\build{d}{d \in \Nmax{\nicefrac{v}{3} + 2}, \Csegmentsize \rem 2d = 0}}\right)
+\end{equation}
+
+This rate is derived from the fact we wish to be able to reconstruct even should almost two-thirds of the $v$ validators be malicious or incapacitated, the 16-bit Galois field on which the erasure-code is based and the desire to support, for simplicity, encoding segments of size $\Csegmentsize$ without padding. The rate is optimal when $\fnecoriginalshards(v) = \nicefrac{v}{3} + 1$, which occurs when $v \in \set{6, 9, 15, 24, 33, 51, 54, 78, 105, 111, 159, 168, 225, 321, 339, 510, 681, 1023}$. It is thus recommended, for efficiency of the data availability system, that validator set sizes be chosen from this set.
```
davxy: Worth noting that for some validator set sizes the divisibility constraint in D(v) forces significantly more redundancy than the fault tolerance threshold requires. The worst case is v=1020, where D(v) drops to 228 (instead of the optimal 341), pushing the rate from ~1:3 to ~1:4.5. This applies to the whole range v in 684..1020, since they all get stuck at D(v)=228 . 

BTW, I'm not sure, if is worth forcing v to be in the optimal set. Maybe the recommendation is sufficient

text/erasure_coding.tex
```diff
@@ -27,24 +34,24 @@ \subsection{Blob Encoding and Recovery}
...
+The original data may be reconstructed with any $\fnecoriginalshards(v)$ of the $v$ resultant items (along with their indices). If the original $\fnecoriginalshards(v)$ items are known then reconstruction is just their concatenation.
 \begin{equation}
   \label{eq:erasurecodinginv}
-  \fnecrecover_{k \in \N}\colon\abracegroup{
-    \protoset{\tuple{\blob[2k], \Nmax{1023}}}_{342} &\to \blob[684k] \\
+  \fnecrecover{v \in \valcount}{k \in \N}\colon\abracegroup{
+    \protoset{\tuple{\blob[2k], \Nmax{v}}}_{\fnecoriginalshards(v)} &\to \blob[2k \cdot \fnecoriginalshards(v)] \\
     \mathbf{c} &\mapsto \begin{cases}
-      \encode{\sq{\build{\mathbf{x}}{\tup{\mathbf{x}, i} \orderedin \sqorderby{i}{\tup{\mathbf{x}, i} \in \mathbf{c}}}}} &\when \set{\build{i}{\tup{\mathbf{x}, i} \in \mathbf{c}}} = \Nmax{342}\\
+      \encode{\sq{\build{\mathbf{x}}{\tup{\mathbf{x}, i} \orderedin \sqorderby{i}{\tup{\mathbf{x}, i} \in \mathbf{c}}}}} &\when \set{\build{i}{\tup{\mathbf{x}, i} \in \mathbf{c}}} = \Nmax{\fnecoriginalshards(v)}\\
```
davxy, line 51: (Small simplification, not strictly related to this PR)
The special case in H.6 for when the available chunks are exactly the first D(v) is not strictly necessary. The general recovery algorithm (R^v) already handles any set of D(v) chunks correctly, including indices 0..D(v)-1. Since we explicitly state that the encoding is systematic, this optimization is not strictly required here. The outcome remains the same.

text/erasure_coding.tex
```diff
@@ -118,9 +125,9 @@ \subsection{Code Word representation}
 m\sub{i} = \tup{m\sub{i, 0} \ldots m\sub{i, 15}}
 \end{equation}
 
-Where $m_{i, 0}$ is the least significant bit of message word $m\sub{i}$. Accordingly we consider the field element $\tilde{m}\sub{i} = \sum^{15}_{j = 0} m\sub{i, j} v\sub{j}$ to represent that message word.
+Where $m_{i, 0}$ is the least significant bit of message word $m\sub{i}$. Accordingly we consider the field element $\widetilde{m\sub{i}} = \sum^{15}_{j = 0} m\sub{i, j} v\sub{j}$ to represent that message word.
 
-Similarly, we assign a unique index to each validator between 0 and 1,022 and we represent validator $i$ with the field element:
+Similarly, we assign a unique index to each validator between 0 and $v - 1$ and we represent validator $i$ with the field element:
```
davxy, line 130: (Again, not strictly related to your changes)
Why do we need to explicitly describe how a validator index is encoded as a codeword?

1. Since this is encoded as a single message word, we could simply refer back to the existing definition.
2. More importantly, why is it necessary to specify how the validator index is encoded here at all? We are erasure coding opaque data (a sequence of m_i).

Perhaps H.10 can be dropped?

text/judgments.tex
```diff
@@ -83,39 +88,40 @@ \subsection{Extrinsic}
...
+    \verdicts &\in \sequence{\tup{\hash, \set{\top, \bot, \none}}} \\
+    \verdicts &\equiv \sq{\build{
+      \tup{\xv¬reporthash, V(\xv¬epochindex, \xv¬judgments)}
     }{
       \tup{\xv¬reporthash, \xv¬epochindex, \xv¬judgments} \orderedin \xtverdicts
-    }}
-\end{align}
+    }} \\
+    V(\xv¬epochindex, \xv¬judgments) &\equiv \begin{cases}
+      \top &\when t = \floor{\twothirds\len{K(\xv¬epochindex)}} + 1 \\
```
davxy: Perhaps we could define **k = K(a)** as in 10.4? Not strictly necessary, just for consistency.

text/reporting_assurance.tex
```diff
@@ -168,26 +173,31 @@ \subsubsection{The Assurances Extrinsic}
...
+    \forall \cX \in \coreindex: \reportspostguarantees\subb{\cX} \equiv {} &\begin{cases}
+      \none &\when p = \none \vee p_\rs¬workreport \in \justbecameavailable \vee {} \\
+      &\phantom{\when} \H_\¬timeslot \ge p_\rs¬timestamp + \Cassurancetimeoutperiod \vee \len{\activeset} \ne \len{\activeset'} \\
+      p &\otherwise
+    \end{cases} \\
+    &\where p = \reportspostjudgement\subb{\cX}
+  \end{aligned}
+\end{equation}
 
+Note that work-reports may time out earlier than usual if the size of the active validator key set $\activeset$ changes.
```
davxy:
```suggestion
Note that all pending work-reports are cleared if the size of the active validator key set $\activeset$ changes.
```

Just personal preference, feel free to ignore :-)

text/work_packages_and_reports.tex
```diff
@@ -112,7 +113,7 @@ \subsection{Packages and Items}\label{sec:packagesanditems}
     &\where S(\wiX \in \workitem) \equiv \len{\wiX_\wi¬payload} + \len{\wiX_\wi¬importsegments}\cdot\Csegmentfootprint + \!\!\!\!\!\!\sum_{\tup{h, l} \in \wiX_\wi¬extrinsics} \!\!\!l
   \end{aligned}\\
   \label{eq:segmentfootprint}
-  &\Csegmentfootprint \equiv \Csegmentsize + 32\ceil{\log_2(\Cmaxpackageimports)}\\
+  &\Csegmentfootprint \equiv \Csegmentsize + 32\ceil{\log_2(\Cmaxpackageexports)}\\
```
davxy, line 116: I imagine we can define a single constant? like `Cmaxpackageios`?

text/work_packages_and_reports.tex
```diff
@@ -184,106 +185,120 @@ \subsubsection{Exporting}
...
+The guarantor must reconstruct imported segments, though this process may in fact be lazy as the Refine function makes no usage of the data until the \emph{fetch} host-call is made. Fetching generally implies that, for each imported segment, erasure-coded chunks are retrieved from enough unique validators (approximately one third) and is described in more depth in appendix \ref{sec:erasurecoding}. (Since we specify systematic erasure-coding, its reconstruction is trivial in the case that the correct validators are responsive.) Chunks must be fetched for both the data itself and for justification metadata which allows us to ensure that the data is correct.
```
davxy, line 287: approx 1:3 in the optimal setup, but maybe not worth annotating here. The reader is already directed to the erasure coding section.

text/reporting_assurance.tex
```diff
@@ -262,26 +274,26 @@ \subsection{Work Report Guarantees}\label{sec:workreportguarantees}
...
-        &s \in \edsignature{(\mathbf{k}\sub{v})_\vk¬ed}{\Xguarantee\concat\blake{\xg¬workreport}}\\
-        &\mathbf{c}\sub{v} = \wrX_\wr¬core \wedge \Crotationperiod(\floor{\nicefrac{\thetime'}{\Crotationperiod}} - 1) \le \xg¬timeslot \le \thetime'\\
-      }\\
-      &k \in \reporters \Leftrightarrow \exists \tup{\xg¬workreport, \xg¬timeslot, \xg¬credential} \in \xtguarantees, \exists \tup{v, s} \in \xg¬credential: k = (\mathbf{k}\sub{v})_\vk¬ed\\
-      &\quad\where \tup{\mathbf{c}, \mathbf{k}} = \begin{cases}
-        \guarantorassignments &\when \displaystyle \ffrac{\thetime'}{\Crotationperiod} = \ffrac{t}{\Crotationperiod} \\
-        \guarantorassignmentsunderlastrotation &\otherwise
-      \end{cases}
+    \end{aligned} : \abracegroup[\,]{
+      &v < \len{\mathbf{k}} \wedge \mathbf{c}\sub{v} = \wrX_\wr¬core < \frac{\len{\activeset'}}{3}\\
```
davxy, line 285:
```suggestion
      &v < \len{\mathbf{k}} \wedge \mathbf{c}\subb{v} = \wrX_\wr¬core < \frac{\len{\activeset'}}{3}\\
```

Subscript notation for sequences

davxy (dismissed):
Looks good, just a few considerations.

text/erasure_coding.tex
zdave-parity: Sure, I elaborated a bit here.

> BTW, I'm not sure, if is worth forcing v to be in the optimal set. Maybe the recommendation is sufficient

I think just having the recommendation is simplest.

text/erasure_coding.tex
zdave-parity: The systematic recovery case I think is just provided to be helpful. Note that the general recovery case has condition "always".

text/erasure_coding.tex
zdave-parity: > More importantly, why is it necessary to specify how the validator index is encoded here at all? We are erasure coding opaque data (a sequence of m_i).

The input to the polynomial is a validator index and is "encoded". FWIW I don't really understand the Cantor basis stuff here, I am just assuming it is correct.

Re improving the wording and definitions in this section, we could do that. In particular it would be nice to make the notation consistent with the rest of the doc. But that can be done in a separate PR...

text/judgments.tex
zdave-parity: Sure, slightly easier to read that way.

text/work_packages_and_reports.tex
zdave-parity: Yeah maybe but there's not really any reason they need to be the same? I think it's sensible to keep the constants separate in case we want to give them different values later.

text/work_packages_and_reports.tex
zdave-parity: Yeah I think it's not worth going into more detail here.

text/reporting_assurance.tex
zdave-parity: Subscript notation is permitted for sequences, from the notation section:

> Sequences are subscriptable, thus a specific item at index i within a sequence s may be denoted s[i], or where unambiguous, s_i

It is used in a few places, presumably because it's a bit shorter. I think it's fine here?

zdave-parity:
Thanks for looking, think I've addressed all the comments.

gavofyork (approved):

PR #517, opened 2026-03-11
Host function gas costs

arkpar:
Introduced formulae and a set of constants for host function gas costs.


text/definitions.tex
```diff
@@ -128,6 +128,7 @@ \subsection{Utilities, Externalities and Standard Functions}
 %  \item[$\mathcal{D}$] Unused.
   \item[$\encode{\dots}$] The octet-sequence encode function. Superscripted ${}^{-1}$ to denote the inverse. See appendix \ref{sec:serialization}.
   \item[$\fnfyshuffle(\dots)$] The Fisher-Yates shuffle function. See equation \ref{eq:suffle}.
+  \item[$\fnmemgas(L, \ell)$] The memory-sized gas function: $\fnmemgas(L, \ell) = \floor{(L \cdot \ell + 1024) / 1024}$. See \ref{sec:gascostcalculation}.
```
gavofyork: We have `\ffrac` for this - we don't use `/` in the GP, `\frac` and `\div`.

text/pvm_invocations.tex
```diff
@@ -852,7 +877,7 @@ \subsection{Accumulate Functions}\label{sec:accumulatefunctions}
   \makecell*[l]{
   $\Omega_T(\gascounter, \registers, \memory, \imXY)$ \\
   \texttt{transfer} = 21 \\
-  $g = 10 + t$} &
+  $g = \CgasT$} &
```
gavofyork: @arkpar this removes the `t` component (the gas limit which the transfer record will execute with). This doesn't seem acceptable since that gas cost must be accounted to avoid the possibility that it can be executed later.

text/definitions.tex
arkpar: Fixed

text/pvm_invocations.tex
arkpar: Right, this should be fixed term + `t`. Fixed.

text/pvm_invocations.tex
```diff
@@ -2,6 +2,13 @@ \section{Virtual Machine Invocations}\label{sec:virtualmachineinvocations}
 
 We now define the three practical instances where we wish to invoke a \textsc{pvm} instance as part of the protocol. In general, we avoid introducing unbounded data as part of the basic invocation arguments in order to minimize the chance of an unexpectedly large \textsc{ram} allocation, which could lead to gas inflation and unavoidable underflow. This makes for a more cumbersome interface, but one which is more predictable and easier to reason about.
 
+\subsection{Gas cost calculation}\label{sec:gascostcalculation}
+
+Memory-sized gas for host-calls is given by a rate $L$ (gas per 1024 octets) and size $\ell$ (octets). The following defines the gas cost for such a size:
+\begin{equation}\label{eq:fnmemgas}
+  \fnmemgas(L, \ell) \equiv \floor{\frac{L \cdot \ell + 1024}{1024}}.
```
gavofyork:
```suggestion
  \fnmemgas(L, \ell) \equiv \ffrac{L \cdot \ell + 1024}{1024}.
```

preamble.tex
```diff
@@ -285,6 +289,75 @@
 \newcommand*{\Cpvminitinputsize}{\mathsf{Z}_I}
 \newcommand*{\Cpvmpagesize}{\mathsf{Z}_P}
 \newcommand*{\Cpvminitzonesize}{\mathsf{Z}_Z}
+\newcommand*{\Cgaschunksize}{\mathsf{G}^c}
```
gavofyork: We don't generally use superscript with lowercase roman characters lest it be confused with the exponentiation operation.

I'd instead replace $\mathsf{G}^c$ with $\mathsf{Z}_G$ and simplify $\mathsf{G}^h$ to just $\mathsf{G}$.

text/pvm_invocations.tex
```diff
@@ -2,6 +2,13 @@ \section{Virtual Machine Invocations}\label{sec:virtualmachineinvocations}
 
 We now define the three practical instances where we wish to invoke a \textsc{pvm} instance as part of the protocol. In general, we avoid introducing unbounded data as part of the basic invocation arguments in order to minimize the chance of an unexpectedly large \textsc{ram} allocation, which could lead to gas inflation and unavoidable underflow. This makes for a more cumbersome interface, but one which is more predictable and easier to reason about.
 
+\subsection{Gas cost calculation}\label{sec:gascostcalculation}
```
gavofyork:
```suggestion
\subsection{Gas Cost Calculation}\label{sec:gascostcalculation}
```

gavofyork (changes requested):
Needs a search-replace.

preamble.tex
arkpar: > Simplify $\mathsf{G}^h$ to just $\mathsf{G}$.

$\mathsf{G}_A$, $\mathsf{G}_I$, $\mathsf{G}_R$, $\mathsf{G}_T$ are already used for gas limits.
Should I use  a free letter such as $\mathsf{M}$ for host function costs?

gavofyork: Well, these are also gas costs which is very similar to gas limits. As long as there's no overlaps in the subscript namespace of $\mathsf{G}$ then I don't see any issue. But I'm also happy to use another letter.

arkpar: There's an overlap. Gas costs use the same subscripts as $\mathsf{\Omega}$ hostcalls. A convention that I think is worth keeping. So e.g. $\mathsf{G}_A$ would be the gas cost for `assign` which conflicts with accumulate gas limit. Going to use $\mathsf{M}$ instead.

gavofyork (dismissed):

text/definitions.tex
```diff
@@ -292,6 +293,78 @@ \subsubsection{Constants}
   \item[$\Cpvminitinputsize = 2^{24}$] The standard \textsc{pvm} program initialization input data size. See equation \ref{sec:standardprograminit}.
   \item[$\Cpvmpagesize = 2^{12}$] The \textsc{pvm} memory page size. See equation \ref{eq:pvmmemory}.
   \item[$\Cpvminitzonesize = 2^{16}$] The standard \textsc{pvm} program initialization zone size. See section \ref{sec:standardprograminit}.
+
+  \medskip
+  \textbf{Host-function gas costs} (appendix \ref{sec:virtualmachineinvocations}). Memory-sized linear terms are given as the rate per 1024 octets; gas for $\ell$ octets at rate $L$ is $\fnmemgas(L, \ell)$ (see \ref{sec:gascostcalculation}). With base cost $c$, total gas is $c + \fnmemgas(L, \ell)$. Page terms are per \textsc{pvm} page. Subscripts match $\Omega$ in I.2. All values are in gas units.
```
zdave-parity: This fake section header is kind of weird? Just make this a separate `\subsubsection` with its own `description` block?

zdave-parity:
```suggestion
These constants are used in the host function gas cost equations in appendix
\ref{sec:virtualmachineinvocations}. The constant term for the host function $\Omega_\square$ is
denoted $\mathsf{M}_{\square,c}$, or simply $\mathsf{M}_\square$ if there are no other terms. The
octet-linear term, if there is one, is denoted $\mathsf{M}_{\square,\ell}$ and specifies gas per
1024 octets. The page-linear term, again optional, is denoted $\mathsf{M}_{\square,p}$ and
specifies gas per page. The gas for a host call with constant term $c$ and octet-linear term $L$
with $\ell$ octets is $c + \fnmemgas(L, \ell)$ (see section \ref{sec:gascostcalculation}).
```

I think this could be quite a bit clearer, perhaps something like this. Would be nice to use a different suffix for page-linear vs octet-linear terms (perhaps p for page-linear terms as I've put in the suggested wording).

text/pvm_invocations.tex
```diff
@@ -287,23 +297,24 @@ \subsection{General Functions}\label{sec:generalfunctions}
 
 \aboverulesep = 1.5mm \belowrulesep = 2mm
 
-\begin{longtable}{p{3.5cm} p{12.5cm}}
+\begin{longtable}{p{5cm} p{11cm}}
   \toprule
-  \thead*{\textbf{Function} \\ \textbf{Identifier} \\ \textbf{Gas usage}} &
+  \thead*{\textbf{Function} \\ \textbf{Identifier} \\ \textbf{Gas usage (}g\textbf{)}} &
```
zdave-parity:
```suggestion
  \thead*{\textbf{Function} \\ \textbf{Identifier} \\ \textbf{Gas usage (}$g$\textbf{)}} &
```

And I guess make the change in the other tables?

text/pvm_invocations.tex
```diff
@@ -287,23 +297,24 @@ \subsection{General Functions}\label{sec:generalfunctions}
...
+  $g = \CgasG$} &
   $\begin{aligned}
     \registers'_7 &\equiv \gascounter'
   \end{aligned}$\\
   \cmidrule(lr){1-1}\cmidrule(lr){2-2}
   \makecell*[l]{
   $\Omega_\Gemini(\gascounter, \registers, \memory, \jam¬blob)$ \\
-  \texttt{grow\_heap} = 1} &
+  \texttt{grow\_heap} = 1 \\
+  $g = \CgasGeminiconst + \max(0, \registers_7 - h) \cdot \CgasGeminilinear$} &
```
zdave-parity: What's going on here? You are defining g twice? AIUI it was omitted here because for better or worse `grow_heap` has custom gas-charging logic.

preamble.tex
```diff
@@ -109,6 +109,10 @@
 \newcommand*{\host}{\hbar}
 \newcommand*{\halt}{\blacksquare}
 \newcommand*{\fault}{\text{\raisebox{6pt}{\rotatebox{180}{\textsf{F}}}}}
+% Zodiac symbols (for $\Omega_\Gemini$, $\Omega_\Aries$, $\Omega_\Taurus$ and gas subscripts)
```
zdave-parity: Why are these needed? The PDF generates fine for me with these definitions removed. Is it to support an old TeX Live version or something?

text/pvm_invocations.tex
```diff
@@ -277,6 +284,9 @@ \subsection{General Functions}\label{sec:generalfunctions}
 Unlike the Accumulate functions in appendix \ref{sec:accumulatefunctions}, these do not mutate an accumulation context. Some, such as $\mathtt{write}$ mutate a service account and both accept and return some $\mathbf{s} \in \serviceaccount$. Others are more general functions, such as $\mathtt{fetch}$ and do not assume any context but have a parameter list suffixed with an ellipsis to denote that the context parameter may be taken and is provided transparently into its result. This allows it to be easily utilized in multiple \textsc{pvm} invocations.
 
 Elements of \textsc{pvm} state are each assumed to remain unchanged by the host-call unless explicitly specified.
+
+Memory-sized gas uses $\fnmemgas$ (see \ref{sec:gascostcalculation}).
```
zdave-parity: This has been inserted between an equation and the text describing the equation. Move this below the equation? I would maybe just inline that whole section here as the section is tiny and only really relevant to the tables here.

text/pvm_invocations.tex
```diff
@@ -324,10 +335,10 @@ \subsection{General Functions}\label{sec:generalfunctions}
     % Current heap pointer.
     \using h &= a + c \\
     % The total gas cost, based on the number of pages that will become writable after this call.
-    \using g &= 100 + (\registers_7 - h) \cdot 10 \\
+    \using g &= \CgasGeminiconst + \max(0, \registers_7 - h) \cdot \CgasGeminilinear \\
```
zdave-parity: FWIW the `\max` here is not necessary as g is only used when `\registers_7 > h`.

text/pvm_invocations.tex
```diff
@@ -339,7 +350,23 @@ \subsection{General Functions}\label{sec:generalfunctions}
   \makecell*[l]{
   $\Omega_Y(\gascounter, \registers, \memory, p, n, \mathbf{r}, i, \overline{\mathbf{i}}, \overline{\mathbf{x}}, \mathbf{i}, \dots)$ \\
   \texttt{fetch} = 2 \\
-  $g = 10$} &
+  $g = \begin{cases}
+    \CgasYprotoacc &\when \registers_{10} = 0 \\
```
zdave-parity: The rendering of this spills over into the next column. Wondering if it would be possible to define this something like `$g = \CgasYc{c} + \fnmemgas(\CgasYl{c}, \len{\mathbf{v}})$` with c = reg10 if <16 else none? Might require duplicating some constants but would be simple and use up way less space here. Guess you could even define the gas constants for this host-call as dictionaries which are looked up by reg10? In any case we should do something to make this fit in the column, it's not very readable as-is.

text/definitions.tex
arkpar: Done

arkpar: Done

text/pvm_invocations.tex
arkpar: Done

text/pvm_invocations.tex
arkpar: A leftover from when I was experimenting moing $g$ to the right column. Removed + removed `max`

preamble.tex
arkpar: Probably a package I was missing. Removed.

text/pvm_invocations.tex
arkpar: Done. `c` is defined in the right column. To be fair there are a few other $g$ definitions that spill a little. Perhaps we should just move all of them to the right column? I'd leave it for a future PR though as it is purely cosmetic.

text/pvm_invocations.tex
arkpar: Done

text/pvm_invocations.tex
zdave-parity: Hmm yeah I see. Also the right column overflows in some places, I assume this is unrelated to this PR. This bit was the most egregious anyway, I think the other overflows aren't that bad. So I would say we address them in a separate PR. Probably just a case of inserting a few manual line breaks.

Re splitting things between left & right columns I think I agree that we should move gas to the right column. I'm not a huge fan of things in the left column using things defined in the right column as I don't think it's super clear. This is done in a few cases for gas, so moving gas to the right column would make these cases clearer IMO.

text/pvm_invocations.tex
```diff
@@ -630,7 +637,7 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \makecell*[l]{
   $\Omega_Z(\gascounter, \registers, \memory, \mathbf{m})$ \\
   \texttt{pages} = 12 \\
-  $g = 10$} &
+  $g =  \begin{cases} \CgasZfreeconst + c \cdot \CgasZfreelinear &\when r = 0 \\ \CgasZallocconst + c \cdot \CgasZalloclinear &\when r \in \set{1, 2} \\ \CgasZsetmodeconst + c \cdot \CgasZsetmodelinear &\when r \in \set{3, 4} \end{cases} $ }&
```
zdave-parity: Missing a gas cost here for invalid `r`.

text/pvm_invocations.tex
```diff
@@ -580,7 +587,7 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \makecell*[l]{
   $\Omega_M(\gascounter, \registers, \memory, \mathbf{m})$ \\
   \texttt{machine} = 9 \\
-  $g = 10$} &
+  $g = \CgasMconst + \fnmemgas(\CgasMlinear, \len{\mathbf{p}})$ }&
```
zdave-parity: Should use `p_Z` instead of `\len{\mathbf{p}}`? `\mathbf{p}` can be `\error`.

text/pvm_invocations.tex
```diff
@@ -528,16 +535,16 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
...
-  \thead*{\textbf{Function} \\ \textbf{Identifier} \\ \textbf{Gas usage}} &
+  \thead*{\textbf{Function} \\ \textbf{Identifier} \\ \textbf{Gas usage (}$g$\textbf{)}} &
   \thead{\textbf{Mutations}} \\
   \cmidrule(lr){1-1}\cmidrule(lr){2-2}
   \endhead
   \makecell*[l]{
   $\Omega_H(\gascounter, \registers, \memory, s, \mathbf{d}, t)$ \\
   \texttt{historical\_lookup} = 7 \\
-  $g = 10$} &
+  $g = \CgasHconst + \fnmemgas(\CgasHlinear, \len{\mathbf{v}})$} &
```
zdave-parity: Probably want to use \registers_{11} here rather than \len{\mathbf{v}}? Or something more complicated if we want to charge less gas when less data is actually fetched. If we use something based on \mathbf{v}, we should (a) account for that being set to some error value and (b) probably charge in proportion to the amount of data requested when it is less than the amount available. If our current impl has cost proportional to the available size rather than the size returned then we should fix this...

Similar comment applies to a bunch of other functions, eg `lookup`, `read`, etc...

text/pvm_invocations.tex
arkpar: Fixed

text/pvm_invocations.tex
arkpar: Introduced an additional constant

text/pvm_invocations.tex
arkpar: Done. Introduced `z` in a way similar to other functions.

text/pvm_invocations.tex
```diff
@@ -285,25 +285,32 @@ \subsection{General Functions}\label{sec:generalfunctions}
   \end{cases}
 \end{align}
 
+Memory-sized gas for host-calls is given by a rate $L$ (gas per 1024 octets) and size $\ell$ (octets). The following defines the gas cost for such a size:
+\begin{equation}\label{eq:fnmemgas}
+  \fnmemgas(L, \ell) \equiv \ceil{\frac{L \cdot \ell}{1024}}.
```
zdave-parity: Stray period at the end of this line.

text/pvm_invocations.tex
```diff
@@ -277,6 +276,7 @@ \subsection{General Functions}\label{sec:generalfunctions}
 Unlike the Accumulate functions in appendix \ref{sec:accumulatefunctions}, these do not mutate an accumulation context. Some, such as $\mathtt{write}$ mutate a service account and both accept and return some $\mathbf{s} \in \serviceaccount$. Others are more general functions, such as $\mathtt{fetch}$ and do not assume any context but have a parameter list suffixed with an ellipsis to denote that the context parameter may be taken and is provided transparently into its result. This allows it to be easily utilized in multiple \textsc{pvm} invocations.
 
 Elements of \textsc{pvm} state are each assumed to remain unchanged by the host-call unless explicitly specified.
+
```
zdave-parity: Remove this inserted line? This affects layout.

text/pvm_invocations.tex
```diff
@@ -659,7 +669,7 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \makecell*[l]{
   $\Omega_K(\gascounter, \registers, \memory, \mathbf{m})$ \\
   \texttt{invoke} = 13 \\
-  $g = 10$} &
+  $g = \CgasK$} &
```
zdave-parity: Should charge for gas used by the inner machine? We could charge up-front for the maximum inner gas usage and then refund any unused gas at the end? Or maybe we could cap inner VM gas to available gas, think that would work without requiring special OOG handling here and would probably be more useful?

text/pvm_invocations.tex
```diff
@@ -691,7 +701,7 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \makecell*[l]{
   $\Omega_X(\gascounter, \registers, \memory, \mathbf{m})$ \\
   \texttt{expunge} = 14 \\
-  $g = 10$} &
+  $g = \CgasX$} &
```
zdave-parity, line 707: For this we assume the cost of freeing memory is covered by the cost of allocating it in the first place? That certainly makes the gas charging simpler.

text/pvm_invocations.tex
```diff
@@ -767,7 +777,7 @@ \subsection{Accumulate Functions}\label{sec:accumulatefunctions}
   \makecell*[l]{
   $\Omega_D(\gascounter, \registers, \memory, \imXY)$ \\
   \texttt{designate} = 17 \\
-  $g = 10$} &
+  $g = \CgasD$} &
   $\begin{aligned}
     \using \sq{o, z} &= \registers\subrange{7}{2} \\
```
zdave-parity, line 785: This now has a count, may want to use that in the gas calc? It's limited though so not necessary.

zdave-parity (commented):
Looks reasonable to me now, just a few final comments.

We could consider relaxing a few of the gas costs in error cases, I see that eg `grow_heap` and `transfer` already do this. I think it's safe and simple as-is though.

text/pvm_invocations.tex
arkpar: Renamed  `g` -> `g_R` in the right column to avoid conflict.
Introduced inner gas cap `g_I`. Remaining inner gas returned from `\Psi` is `g_I'`

text/pvm_invocations.tex
arkpar: Benchmarking shows that there is no linear factor for `free` in polkajam, so I did not introduce it.

text/pvm_invocations.tex
arkpar: Removed

text/pvm_invocations.tex
arkpar: Removed

text/pvm_invocations.tex
arkpar: Done

text/pvm_invocations.tex
```diff
@@ -659,16 +668,20 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
   \makecell*[l]{
   $\Omega_K(\gascounter, \registers, \memory, \mathbf{m})$ \\
   \texttt{invoke} = 13 \\
-  $g = 10$} &
+  $g = \CgasK + g_I - g_I'$} &
```
gavofyork: I don't think this will work since, due to (B.20), the purpose of $g$ is to gate the function on available gas, thus $g$ must be trivially computable prior to instruction execution.

The way to handle invoke is to set $g$ as the maximum possible gas to be deducted based purely on the operands and then explicitly set $\varrho'$ to something other than $\varrho - g$ as needed.

gavofyork (changes requested):
Invoke needs correcting I think...

text/pvm_invocations.tex
arkpar: After @gavofyork's comment here: https://github.com/gavofyork/graypaper/pull/517#discussion_r3336869593

1. Tuple decoding error sets `g_R` to zero so that it is always a number and `g` definition works. Also updated panic condition to use `\mathbf{w}`
2. Introduced explicit `\gascounter'` override

koute:
We also need to set the default gas when an unknown hostcall is called, e.g. on line 53 you have (also on line 109 and 212, which all could just use the same constant):

```
\multicolumn{2}{l}{\also \gascounter' = \gascounter - 10}
```

We could either make this cheap (i.e. make it cost as much as calling a hostcall which does nothing, so just copy the same gas as we charge for `gas`, I guess), *or* we could make it more expensive - expensive enough to cover e.g. the non-GP `log` hostcall, in which case it'd make it safe to call the logger without having to explicitly define it in the GP or need to disable it in production.

arkpar:
Added a constant for unknown hostcall wich is reasonably expensive (1000 gas)

text/pvm_invocations.tex
```diff
@@ -677,8 +686,12 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
       \end{cases}\\
       \mathbf{m}^*\subb{n}_{\pg¬gaschargedflag} = \pg¬gaschargedflag'\\
     \end{cases}\\
+    \gascounter' &\equiv \begin{cases}
+      \gascounter - \CgasK &\when \mathbf{w} = \error \vee n \not\in \mathbf{m} \\
```
davxy:
```suggestion
      \gascounter - \CgasK &\when \mathbf{w} = \error \vee n \not\in \keys{\mathbf{m}} \\
```

text/pvm_invocations.tex
zdave-parity: FWIW I think this was fine? The implementation would charge `\Cgask` up-front, and then whatever gas was actually used by the inner invoke afterwards. It would not be possible for this second charge to underflow gas as it would always be <= the available gas, due to the inner VM's gas limit being `min(g_R, \gascounter - \CgasK)`. Edit: Guess you would also need a `max(..., 0)` on this.

zdave-parity (dismissed):

text/pvm_invocations.tex
zdave-parity: It is surely linear in the number of pages allocated by the VM? I guess it wouldn't make much sense to charge for this here as we have to dealloc the pages for free when the outer VM exits anyway. So any cost of doing this needs to be paid up-front as part of allocation or something. Maybe the extra cost is so small that it can just be ignored.

koute: > I guess it wouldn't make much sense to charge for this here as we have to dealloc the pages for free when the outer VM exits anyway. So any cost of doing this needs to be paid up-front as part of allocation or something.

Yes. Since the user can just not call `expunge` we can't really depend on it to charge for deallocating resources. This cost needs to be accounted for when the VM is created/when pages are allocated, so the cost of `expunge` would mostly be just a small constant cost, probably just slightly more expensive than how much a hostcall which does nothing costs.

text/pvm_invocations.tex
```diff
@@ -677,8 +686,12 @@ \subsection{Refine Functions}\label{sec:refinefunctions}
       \end{cases}\\
       \mathbf{m}^*\subb{n}_{\pg¬gaschargedflag} = \pg¬gaschargedflag'\\
     \end{cases}\\
+    \gascounter' &\equiv \begin{cases}
+      \gascounter - \CgasK &\when \mathbf{w} = \error \vee n \not\in \keys{\mathbf{m}} \\
+      \gascounter - g + g_R' &\otherwise
```
zdave-parity, line 691: Just realised this is not right; we should only refund unused gas in the case where there was enough gas to invoke the inner VM? If `\gascounter < g`, `\gascounter'` should simply be `\gascounter - g`.

arkpar: Fixed. One thing of note when calling `invoke` for a non-existing machine.  If there's less than \CgasK + g_R actual gas left, we'll charge the full g and end with OOG. However if there's over  \CgasK + g_R gas left, we'll get WHO and only charge \CgasK?

zdave-parity: Yeah this is weird. Can we just set `\gascounter' = \gascounter - g` in the second case here as well?

zdave-parity: So put here:

    \gascounter - g &\when \mathbf{w} = \error \vee n \not\in \keys{\mathbf{m}} \vee \gascounter < g \\

zdave-parity (dismissed):

text/pvm_invocations.tex
arkpar: Done

zdave-parity (approved):

gavofyork (approved):

PR #520, opened 2026-03-22
Constrain preimage length parameter to bloblength (< 2^32) in preimage-related host calls

sierkov:
### Background

The four preimage-related host calls (```solicit```, ```forget```, ```query```, ```provide```) accept the preimage length as z ∈ pvmreg = N_{2^{64}}. However, the preimage lookup dictionary ((9.3) in GP 0.7.2) stores the size as bloblength = N_L = N_{2^{32}}. The host call definitions are therefore inconsistent with the account type they operate on.

This inconsistency propagates into the storage footprint and threshold balance formulae ((9.8) in GP 0.7.2).
 A single request entry with z = 2^{64} - 1 contributes 81 + z = 2^{64} + 80 to the sum — a value that cannot be represented in the declared type N_{2^{64}}. Even for values of z where the storage footprint fits in 64 bits, the subsequent threshold balance computation multiplying it with B_L (currently 1 but can be increased in the future) could exceed N_{2^{64}}, requiring 128-bit intermediate arithmetic or overflow detection that may not be available in all implementation languages.

For provide specifically, z is also used to read memory. When z >= 2^{32} the memory read would exceed the 32-bit PVM address space and produce a panic exit — the wrong result code for an invalid argument. An explicit check is therefore needed to return HUH rather than panic.

### Proposal

All four calls return HUH if z >= 2^{32}, checked before any memory access or state lookup.

### Benefits

1. Fixes the type inconsistency: with z < 2^{32}, the octets sum stays within N_{2^{64}} for a realistic number of entries, making the declared type correct.
2. Corrects the error code for provide: large z currently produces panic; the explicit check returns the more appropriate HUH.
3. Reduces cross-implementation divergence: an explicit bound removes a source of inconsistency between implementations that may handle the large-z case differently.
4. Aligns with the existing definitions: the preimage lookup dictionary key is already defined as (hash, bloblength) — the host calls should enforce the same constraint.

### DIFF
--- a/text/pvm_invocations.tex
+++ b/text/pvm_invocations.tex
@@ -922,7 +922,7 @@ Other than the gas-counter which is explicitly defined, elements of \textsc{pvm}
     \end{cases} \\
     \tup{\execst', \registers'_7, \registers'_8} &\equiv \begin{cases}
       \tup{\continue, \mathtt{HUH}, \registers_8} &\when z \not\in \bloblength \\
-      \tup{\panic, \registers_7, \registers_8} &\when h = \error \\
+      \tup{\panic, \registers_7, \registers_8} &\otherwhen h = \error \\
       \tup{\continue, \mathtt{NONE}, 0} &\otherwhen \mathbf{a} = \error \\
       \tup{\continue, 0, 0} &\otherwhen \mathbf{a} = \sq{} \\
       \tup{\continue, 1 + 2^{32}x, 0} &\otherwhen \mathbf{a} = \sq{x} \\
@@ -949,7 +949,7 @@ Other than the gas-counter which is explicitly defined, elements of \textsc{pvm}
     \end{cases} \\
     \tup{\execst', \registers'_7, \imX'_\im¬self} &\equiv \begin{cases}
       \tup{\continue, \mathtt{HUH}, \imX_\im¬self} &\when z \not \in \bloblength \\
-      \tup{\panic, \registers_7, \imX_\im¬self} &\when h = \error \\
+      \tup{\panic, \registers_7, \imX_\im¬self} &\otherwhen h = \error \\
       \tup{\continue, \mathtt{HUH}, \imX_\im¬self} &\otherwhen \mathbf{a} = \error \\
       \tup{\continue, \mathtt{FULL}, \imX_\im¬self} &\otherwhen \mathbf{a}_\sa¬balance < \mathbf{a}_\sa¬minbalance \\
       \tup{\continue, \mathtt{OK}, \mathbf{a}} &\otherwise \\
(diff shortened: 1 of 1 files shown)


gavofyork:
With a sensible total issuance of balance tokens, it should not be possible to have a balance greater than the threshold balance implied by even a single storage or lookup item of $2^{32} - 1$ bytes.

Therefore in the case of `solicit` we can safely assume that if the $\phi'_7$ is not `FULL` (and the account state is unchanged), then $z < 2^{32}$. Since lookup entries can only be inserted with `solicit`, it is consistent. In the case of others, no entries with $z \ge 2^{32}$ can have existed since `solicit` alone can insert entries into the account's lookup mapping.

Regarding `solicit`, I'd agree that the inconsistency in the definition of $\mathbf{a}$ is questionable and it is unclear what value $\mathbf{a}$ should take, leaving the `HUH` branch unevaluable. This is an issue and my preferred fix (which adds no further complexity to the definitions) is to simply reverse the order of `HUH` and `FULL` errors, meaning that overly large values of $z$ will trigger the `FULL` branch, circumventing the need to have a definite value for $\mathbf{a}$.

sierkov:
I fully agree that this issue does not arise if a service’s balance can never exceed 2^32 - 1. This may already have been discussed internally, and I may be missing some context, but I’d like to offer three observations from the perspective of an independent implementer working solely from the GP text.

### Service balances exceeding 2^32 - 1 seem plausible

GP defines the balance type as 64-bit. The discussion section describes each of the 1,023 validators as having 8 TB of secondary storage. Appendix H specifies erasure coding with a 342:1,023 reconstruction threshold, meaning unique data capacity is 342 × 8 TB ≈ 2.7 PB, or roughly 2 PB after accounting for overhead. At the current rate of B_L = 1 token per byte, funding that storage would require on the order of 2^51 tokens in circulation, making individual service balances well above 2^32 plausible.

With that context, consider a data-intensive service such as a public key registry requiring 256 bytes per key (key, email, validity period, etc.). Serving more than 2^24 users would already require more than 4 GB of preimage storage, and thus a balance exceeding 2^32.

### The type inconsistency has no clean resolution without an explicit guard

Inserting a key with z > 2^32 - 1 into the preimage lookup dictionary is undefined. However, the solicit host call must handle the case where a key is not found, which should apply to any key outside the domain of bloblength.

This not only makes the HUH branch unreachable, but also forces implementations into undefined behavior, such as capping the value at 2^32 - 1, truncating it to the lowest 32 bits, or storing preimage length as a 64-bit value, creating a potential consensus risk.

### On the FULL-before-HUH swap

Swapping the cases makes the behavior well-defined when a service balance can never exceed 2^32 - 1, but it also produces a misleading error code by describing something other than the true root cause: FULL signals insufficient balance, not an invalid argument.

In my view, adding an explicit z ∉ bloblength → HUH guard not only resolves the type inconsistency directly, but also allows for larger service balances and makes the error semantics more straightforward: FULL means insufficient balance, and HUH means an invalid parameter.

gavofyork (approved):

PR #513, opened 2026-02-17
Introduce hard-limits on the size of dispute extrinsics, and a rationale for it.

alxmirap:
Edit: this Pr was closed and superseded by a rebased version in https://github.com/gavofyork/graypaper/pull/525

Edit: 

Adds another change:

- removes the minimum culprits and faults to be included in an extrinsic, giving an explanation why.

This removal is caused by the impossibility of reporting the same offender in more than one extrinsic, and the possibility that one validator can indeed cast more than one vote liable to cause offence.

PR #523, opened 2026-05-07
Removes offense minimums on top of fork branch

alxmirap:
This subsumes the previous pull-request #513 and adds another change:
- removes the minimum culprits and faults to be included in an extrinsic, giving an explanation why.

This removal is caused by the impossibility of reporting the same offender in more than one extrinsic, and the possibility that one validator can indeed cast more than one vote liable to cause offence.

Although this tension was not introduced by the limit on the size of offender extrinsics and the possible necessity of creating these after the verdict (introduced by #513), it touches the same part of the text and behaves like a whole.


alxmirap:
Redundant due to #513

PR #524, opened 2026-05-21
Propose change of extrinsic hash definition to allow checking for individual extrinsic preimages.

cheme:
This PR modify the way extrinsic preimages are hashed to allow verifying a single preimage without having to download all preimages. (it simply replace preimage data by its blake hash).

cc\ @zdave-parity


gavofyork (dismissed):
Looks harmless enough to me

text/header.tex
zdave-parity: Guess we should say "possibility of reports and preimages" rather than just reports.

zdave-parity (dismissed):
Looks sensible to me.

text/header.tex
cheme: Indeed, I added your suggestion.

zdave-parity (approved):

gavofyork (approved):

PR #525, opened 2026-05-25
Introduce hard-limits on the size of dispute extrinsics, and remove culprits requirement

alxmirap:
Introduces hard-limits on the size of dispute extrinsics and a rationale for it.
Removes the requirement of a minimum number of culprits per verdict for a Bad report. This is caused by the impossibility of reporting the same offender in more than one extrinsic, and the possibility that one validator could provide guarantees for a second bad report after being already identified as an offender.

This supersedes PR #513.


zdave-parity:
I've set the title to what it was before.

text/judgments.tex
```diff
@@ -23,21 +23,21 @@ \subsection{The State}
...
+The disputes extrinsic $\xtdisputes$ is functional grouping of three otherwise independent extrinsics. It comprises \emph{verdicts} $\xtverdicts$, \emph{culprits} $\xtculprits$, and \emph{faults} $\xtfaults$. Verdicts are a compilation of judgments coming from exactly two-thirds plus one of either the active validator set or the previous epoch's validator set, \ie the Ed25519 keys of $\activeset$ or $\previousset$. Culprits and faults are proofs of the misbehavior of one or more validators, respectively either by guaranteeing a work-report found to be invalid, or by signing a judgment found to be in contradiction to a work-report's validity. Both of these are considered a kind of \emph{offense}. Formally:
 \begin{equation}
   \begin{aligned}
     \xtdisputes &\equiv \tuple{\xtverdicts, \xtculprits, \xtfaults} \\
-    \where \xtverdicts &\in \sequence{\tuple{
+    \where \xtverdicts &\in \sequence[:\Cmaxextrinsicverdicts]{\tuple{
       \hash,
       \ffrac{\thetime}{\Cepochlen} - \N_2,
-      \sequence{\tuple{
+      \sequence[\floor{\twothirds\valcount} + 1]{\tuple{
```
zdave-parity: Revert this change? Not sure it is well-defined as-is (`\valcount` is a set) and it's not necessary. The 2/3+1 requirement is now specified below, as it depends on the length of the validator set that is selected in the verdict.

FWIW I also intentionally moved the "exactly two-thirds plus one" text from above to the paragraph below, so that the equations would correspond to the wording above them.

text/judgments.tex
```diff
@@ -162,6 +154,26 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+There are some logical consequences derived from the composition of this extrinsic: any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits. 
+Faults and culprits must never list keys that have already been included in the punish-set following some previously seen offence (equations \ref{eq:culprits} and \ref{eq:faults}), and so although logically these minimum offenses must exist for each corresponding verdict, they may not be actually included in a faults or culprits extrinsic: good verdicts should only be generated when there is at least one associated new fault offender; bad verdicts should always be generated even when they don't have any new associated culprit offender since we need to exclude bad reports.
+
+Formally:
+\begin{align}
+  \forall \tup{\¬reporthash, \floor{\twothirds\valcount} + 1} \in \verdicts &:
```
zdave-parity: `\floor{\twothirds\valcount} + 1` should be `\top` now.

text/judgments.tex
```diff
@@ -162,6 +154,26 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+There are some logical consequences derived from the composition of this extrinsic: any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits. 
+Faults and culprits must never list keys that have already been included in the punish-set following some previously seen offence (equations \ref{eq:culprits} and \ref{eq:faults}), and so although logically these minimum offenses must exist for each corresponding verdict, they may not be actually included in a faults or culprits extrinsic: good verdicts should only be generated when there is at least one associated new fault offender; bad verdicts should always be generated even when they don't have any new associated culprit offender since we need to exclude bad reports.
```
zdave-parity:
```suggestion
Faults and culprits must never list keys that have already been included in the punish-set following some previously reported offence (equations \ref{eq:culprits} and \ref{eq:faults}), and so although logically these minimum offenses must exist for each corresponding verdict, it may not be possible to include them in the extrinsic. Good verdicts should only be generated when there is at least one associated new fault offender, as they otherwise would have no effect. Bad verdicts should always be generated even when they don't have any new associated culprit offender since we need to exclude bad reports.
```

text/judgments.tex
```diff
@@ -162,6 +154,26 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+There are some logical consequences derived from the composition of this extrinsic: any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits.
```
zdave-parity:
```suggestion
Any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits. 
```

The "derived from the composition of this extrinsic" bit doesn't make sense to me; these are more logical consequences of how auditing works?

text/judgments.tex
```diff
@@ -162,6 +154,26 @@ \subsection{Extrinsic}
...
+Faults and culprits must never list keys that have already been included in the punish-set following some previously seen offence (equations \ref{eq:culprits} and \ref{eq:faults}), and so although logically these minimum offenses must exist for each corresponding verdict, they may not be actually included in a faults or culprits extrinsic: good verdicts should only be generated when there is at least one associated new fault offender; bad verdicts should always be generated even when they don't have any new associated culprit offender since we need to exclude bad reports.
+
+Formally:
+\begin{align}
+  \forall \tup{\¬reporthash, \floor{\twothirds\valcount} + 1} \in \verdicts &:
+    \exists \tup{\¬reporthash, \dots} \in \xtfaults
+\end{align}
+
+Although this should not be expected in practice, in theory it is possible that the number of offenses could be large (up to $\onethird \valcount$ faults and up to three culprits for each bad verdict). In order not to burden block creation overmuch, we specify practical limits to the number of verdicts and offenses which can be included in a single block.
```
zdave-parity: Again `\onethird \valcount` doesn't really make sense, `\valcount` is the set of possible validator counts.

text/judgments.tex
```diff
@@ -162,6 +154,26 @@ \subsection{Extrinsic}
...
+
+Formally,
```
zdave-parity: These equations are redundant as the limits are specified in the first equation in this section. Probably all this explanatory text about limits is not really necessary either, though it's also probably harmlesss and it seems accurate to me.

text/judgments.tex
alxmirap: Thanks. Although I tried to isolate my changes as best I could, I did not notice the change in the `exact` wording. I am going to revert both of those changes.

text/judgments.tex
alxmirap: Thanks. Fixing.

text/judgments.tex
alxmirap: Ok, rewording to use "$\onethird$ of validators"

text/judgments.tex
alxmirap: I'll remove the equations then. I'm keen to leave the text in, though, but make it point to equation 10.2

text/judgments.tex
```diff
@@ -162,6 +155,19 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+Any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits. 
+Faults and culprits must never list keys that have already been included in the punish-set following some previously reported offence (equations \ref{eq:culprits} and \ref{eq:faults}), and so although logically these minimum offenses must exist for each corresponding verdict, it may not be possible to include them in the extrinsic. Good verdicts should only be generated when there is at least one associated new fault offender, as they otherwise would have no effect. Bad verdicts should always be generated even when they don't have any new associated culprit offender since we need to exclude bad reports.
```
gavofyork: Is this meant to be another paragraph? If so it needs a newline before the last one.

text/judgments.tex
```diff
@@ -162,6 +155,19 @@ \subsection{Extrinsic}
...
+
+Formally:
+\begin{align}
+  \forall \tup{\¬reporthash, \top} \in \verdicts &:
+    \exists \tup{\¬reporthash, \dots} \in \xtfaults
+\end{align}
+
+Although this should not be expected in practice, in theory it is possible that the number of offenses could be large (up to $\onethird$ of validators as faults and up to three culprits for each bad verdict). In order not to burden block creation overmuch, we specify practical limits to the number of verdicts and offenses which can be included in a single block as can be seen in equation \ref{eq:disputesextrinsics}.
```
gavofyork:
```suggestion
Although this should not be expected in practice, in theory it is possible that the number of offenses could be large (up to $\onethird$ of validators as faults and up to three culprits for each bad verdict). In order not to over burden block creation, we specify practical limits to the number of verdicts and offenses which can be included in a single block as can be seen in equation \ref{eq:disputesextrinsics}.
```

text/judgments.tex
```diff
@@ -162,6 +155,19 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+Any verdict containing solely valid judgments implies the existence of at least one validator's vote against it that must have triggered the dispute, which in turn implies a fault; and any verdict containing solely invalid judgments implies the existence of at least two guarantees signed for that report, which in turn implies the existence of at least two culprits.
```
gavofyork: This seems like a less clear rewrite of my previous paragraph: talking about "validators vote" as a separate consideration, but strictly implying, a "fault" is nonsense. Stick with the original wording.

gavofyork:
@alxmirap needs a rebase.

text/judgments.tex
```diff
@@ -162,6 +155,16 @@ \subsection{Extrinsic}
   \end{split}
 \end{align}
 
+There are some constraints placed on the composition of this extrinsic: any verdict containing solely valid judgments implies the same report having at least one valid entry in the faults sequence $\xtfaults$. Good verdicts should only be reported when there is at least one associated new fault offender, as they otherwise would have no effect. Bad verdicts should always be reported.
```
zdave-parity:
```suggestion
Any verdict containing solely valid judgments implies the same report having at least one valid entry in the faults sequence $\xtfaults$. Formally:
```

The second sentence seems to just be a rewording of the first. Also wondering what the reason is for moving this from where it was before?

text/judgments.tex
```diff
@@ -60,12 +61,12 @@ \subsection{Extrinsic}
 
 Offender signatures must be similarly valid and reference work-reports with judgments and may not report keys which are already in the punish-set:
 \begin{align}
-  \forall \tup{\xc¬reporthash, \xc¬offenderindex, \xc¬signature} &\in \xtculprits : \bigwedge \abracegroup{
+  \label{eq:culprits}\forall \tup{\xc¬reporthash, \xc¬offenderindex, \xc¬signature} &\in \xtculprits : \bigwedge \abracegroup{
```
zdave-parity: These labels not referenced anywhere now.

text/judgments.tex
```diff
@@ -162,6 +155,16 @@ \subsection{Extrinsic}
...
 
+There are some constraints placed on the composition of this extrinsic: any verdict containing solely valid judgments implies the same report having at least one valid entry in the faults sequence $\xtfaults$. Good verdicts should only be reported when there is at least one associated new fault offender, as they otherwise would have no effect. Bad verdicts should always be reported.
+
+Formally:
+\begin{align}
+  \forall \tup{\¬reporthash, \top} \in \verdicts &:
+    \exists \tup{\¬reporthash, \dots} \in \xtfaults
+\end{align}
+
+Although this should not be expected in practice, in theory it is possible that the number of offenses could be large (up to $\onethird$ of validators as faults and up to three culprits for each bad verdict). In order not to over burden block creation, we specify practical limits to the number of verdicts and offenses which can be included in a single block as can be seen in equation \ref{eq:disputesextrinsics}.
```
zdave-parity: Would probably be clearer to put this text by the equations that formalise it. Personally I would just remove it as I don't think it adds much.

gavofyork:
@alxmirap needs a rebase in addition to those comments resolving.

gavofyork (approved):

davxy (approved):

PR #528, opened 2026-05-27
Remove unused arguments from refine hostcalls

koute:
This PR removes unused arguments from a couple of refine hostcalls (i.e. if a hostcall doesn't access inner PVM state then it doesn't need it passed; same for exports).


zdave-parity (approved):

gavofyork (approved):

zdave-parity:
It has occurred to me that this is broken... the `Omega` functions are expected to pass the context through even if they do not do anything with it. The context is implicitly passed through unchanged when not otherwise stated. Or at least that's supposed to happen, it seems that is missing from the pass-through equation. Will make a PR to fix on Monday...

zdave-parity:
> It has occurred to me that this is broken... the `Omega` functions are expected to pass the context through even if they do not do anything with it. The context is implicitly passed through unchanged when not otherwise stated. Or at least that's supposed to happen, it seems that is missing from the pass-through equation. Will make a PR to fix on Monday...

Reverted as part of #533.

PR #530, opened 2026-05-28
Add dedicated hostcall for inner PVM module compilation

koute:
This PR adds a dedicated hostcall to compile inner PVM modules.

* The new `compile` hostcall which can compile a new PVM module from outer PVM's memory, *or* directly from a preimage in the database (this avoids the need to first fetch and copy the module's bytecode into the outer PVM's memory; it's essentially a `historical_lookup` + compiling from memory call merged into one)
* `machine` now accepts a handle returned by the `compile` hostcalls (this makes it much cheaper to instantiate multiple VMs provided they use the same module)
* `expunge` hostcall can now also destroy a compiled module handle, and accepts an extra argument which determines the type of handle that was passed
* There's a maximum limit of 128 modules that can be kept in memory at the same time (this is subject to change if necessary)
* The compiled modules are reference counted, so if there exists a live VM instance which uses a given PVM module then that module is also considered alive, even if it was `expunge`d

Arguments to the `compile` hostcall are as follows:
* φ7 (`m`) -> mode of operation, with `0` being from memory and `1` from a preimage
* φ8 (`p`) -> pointer to the blob (memory mode) or a hash of the blob (preimage mode)
* φ9 (`l`) -> length of the blob (memory mode) or maximum length of the blob used (preimage mode)
* φ10 (`f`) -> offset in the blob (preimage mode only)
* φ11 (`i`) -> service ID (preimage mode only)

Arguments to the `expunge` hostcall are:
* φ7 (`m`) -> mode of operation, with `0` being an expunge of a VM instance, and `1` being an expunge of a program module
* φ8 (`n`) -> handle to expunge

[Rendered PDF](https://github.com/koute/misc/blob/master/graypaper-compile-hostcalls-v4.pdff).


gavofyork:
There's no precedent on using multi-character subscripts in the $$\Omega$ formulations. I'd prefer to keep to single characters.

gavofyork:
I also wonder how bad it would be to have the two `compile_` variants be folded into a unified `compile` host-call? This is the convention thus far. I'd really like to keep the conceptual dictionary as minimal as possible.

koute:
Sure, I'll change the subscripts to be single letter.

I can also try to unify both `compile_` hostcalls into one, and we can see how it looks like.

If you'd like we could also unify `expunge` with `drop` by having only `expunge`, and pass the type of the handle as an argument.

koute:
I've merged the two hostcalls into one. I've also merged `drop` into `expunge` while I was at it (let me know if you'd like them separate, in which case I can revert the change).

Also updated the PR description to describe the current state of the PR.

gavofyork:
Yeah, looks good. Needs resolving...

koute:
Aligned with the newest changes on `main`.

gavofyork:
Will need re-resolving again:) @koute

koute:
Done. Should aligned with what's on `main` now.

preamble.tex
```diff
@@ -450,11 +453,14 @@
 
 %%%%%%%%%%%%
 % Named types (blackboard)
-\newcommand*{\innerpvm}{\mathbb{I}}
+\newcommand*{\pvm¬instance}{\mathbb{I}}
+\newcommand*{\pvm¬module}{\blob}
```
zdave-parity: I think `\pvm¬module` and `\pvm¬module¬map` should go in the "PVM-SPECIFIC ITEMS" section? Bit confusing to have them here as this section is otherwise just named types plus the names of their fields.

text/pvm_invocations.tex
```diff
@@ -463,28 +464,57 @@ \subsubsection{Refine Functions}\label{sec:refinefunctions}
...
     \using \mathbf{p} &= \begin{cases}
-      \memory\subrange{p_O}{p_Z} &\when \Nrange{p_O}{p_Z} \subseteq \readable{\memory} \\
-      \error &\otherwise
+      \memory\subrange{p}{l} &\when m = 0 \wedge \Nrange{p}{l} \subseteq \readable{\memory} \\
+      \error &\otherwhen m = 0 \vee (m = 1 \wedge \Nrange{p}{32} \not\subseteq \readable{\memory}) \\
+      \mathbf{h}\subrange{f^*}{l^*} &\otherwhen m = 1 \wedge \mathbf{a} \ne \none \wedge \mathbf{h} \ne \none \\
+      \multicolumn{2}{l}{\where \mathbf{h} = \histlookup(\mathbf{a}, t, \memory\subrange{p}{32}),\ f^* = \min(f, \len{\mathbf{h}}),\ l^* = \min(l, \len{\mathbf{h}} - f^*)} \\
+      \none &\otherwise
     \end{cases} \\
+    \using \mathbf{a} &= \begin{cases}
```
zdave-parity: Move this above `mathbf{p}` as that depends on this?

text/pvm_invocations.tex
```diff
@@ -463,28 +464,57 @@ \subsubsection{Refine Functions}\label{sec:refinefunctions}
...
+      \mathbf{d}\subb{s} &\when i = 2^{64} - 1 \wedge s \in \keys{\mathbf{d}} \\
```
zdave-parity: FWIW you don't need the `\in \keys` checks as `d[k]` is defined to be `\none` when `k` is not in `d`.

text/pvm_invocations.tex
```diff
@@ -463,28 +464,57 @@ \subsubsection{Refine Functions}\label{sec:refinefunctions}
...
+      \mathbf{d}[i] &\when i \in \keys{\mathbf{d}} \\
+      \none &\otherwise
+    \end{cases} \\
+    \using \mathbf{H} &= \keys{\pvm¬module¬map} \cup \set{\mathbf{m}\subb{n}_\pg¬handle : n \in \keys{\mathbf{m}}} \\
+    \using k &= \min(n \in \N : n \not\in \mathbf{H}) \\
+    \tup{\execst', \registers'_7, \pvm¬module¬map'} &\equiv \begin{cases}
+      \tup{\continue, \mathtt{FULL}, \pvm¬module¬map} &\when \len{\mathbf{H}} \ge 128 \\
+      \tup{\panic, \registers_7, \pvm¬module¬map} &\otherwhen \mathbf{p} = \error \\
+      \tup{\continue, \mathtt{NONE}, \pvm¬module¬map} &\otherwhen \mathbf{p} = \none \\
```
zdave-parity: A bit odd that an invalid `m` value returns `NONE` instead of `HUH`.

preamble.tex
```diff
@@ -337,6 +336,10 @@
 \newcommand*{\CgasTaurus}{\mathsf{M}_{\Taurus}}
 \newcommand*{\CgasAriesconst}{\mathsf{M}_{\Aries,c}}
 \newcommand*{\CgasArieslinear}{\mathsf{M}_{\Aries,\ell}}
+\newcommand*{\CgasGammamem}{\mathsf{M}_{\Gamma,\mathrm{m}}}
```
zdave-parity:
```suggestion
\newcommand*{\CgasGammamemconst}{\mathsf{M}_{\Gamma,\mathrm{m},c}}
```

(For consistency with the other constants)

zdave-parity (dismissed):
LGTM, few trivial nits.

preamble.tex
```diff
@@ -844,6 +841,9 @@
 %%%%%%%%%%%%
 % PVM-SPECIFIC ITEMS
 
+\newcommand*{\pvm¬instance}{\mathbb{I}}
```
zdave-parity: `pvm¬instance` should stay where it was? The pattern in the "Named types" section is that you have the "named type" (eg `pvm¬instance`) followed by the names of the type's fields (for `pvm¬instance` these are `pg¬code`, `pg¬ram`, etc).

koute: Hm, well, `\pvm¬instance` (`\mathbb{I}`) is the type of the instance, and `\pvm¬module` (`\mathbb{B}`) is the type of the module. To me it makes sense to keep them together? I can split them, but that seems to be somewhat confusing, considering they're inherently related (you create an instance from a module).

preamble.tex
zdave-parity: Mostly I meant the ",c" on the end. Same comment applies to `\CgasGammadb`.

koute: Ah, sorry! I brainfarted and missed that.

preamble.tex
zdave-parity: I guess as `\pvm¬module` is a type it makes sense in the "Named types" section, but I found it confusing placed between the definition of `\pvm¬instance` and its fields. Anyway it doesn't really matter that much so I'll shut up now!

koute: Heh, okay, well, in general I'm fine with whatever you prefer. (:

zdave-parity (dismissed):

gavofyork (dismissed):

text/pvm_invocations.tex
```diff
@@ -463,28 +464,55 @@ \subsubsection{Refine Functions}\label{sec:refinefunctions}
...
+      \mathbf{h}\subrange{f^*}{l^*} &\otherwhen m = 1 \wedge \mathbf{a} \ne \none \wedge \mathbf{h} \ne \none \\
+      \multicolumn{2}{l}{\where \mathbf{h} = \histlookup(\mathbf{a}, t, \memory\subrange{p}{32}),\ f^* = \min(f, \len{\mathbf{h}}),\ l^* = \min(l, \len{\mathbf{h}} - f^*)} \\
+      \none &\otherwise
     \end{cases} \\
+    \using \mathbf{H} &= \keys{\pvm¬module¬map} \cup \set{\mathbf{m}\subb{n}_\pg¬handle : n \in \keys{\mathbf{m}}} \\
+    \using k &= \min(n \in \N : n \not\in \mathbf{H}) \\
+    \tup{\execst', \registers'_7, \pvm¬module¬map'} &\equiv \begin{cases}
+      \tup{\continue, \mathtt{FULL}, \pvm¬module¬map} &\when \len{\mathbf{H}} \ge 128 \\
+      \tup{\panic, \registers_7, \pvm¬module¬map} &\otherwhen \mathbf{p} = \error \\
+      \tup{\continue, \mathtt{HUH}, \pvm¬module¬map} &\otherwhen \mathbf{p} = \none \vee \deblob(\mathbf{p}, 0) = \error \\
```
davxy, line 505:
```suggestion
      \tup{\panic, \registers_7, \pvm¬module¬map} &\otherwhen \mathbf{p} = \error \\
      \tup{\continue, \mathtt{NONE}, \pvm¬module¬map} &\otherwhen m = 1 \wedge \mathbf{p} = \none \\
      \tup{\continue, \mathtt{HUH}, \pvm¬module¬map} &\otherwhen \mathbf{p} = \none \vee \deblob(\mathbf{p}, 0) = \error \\
```

Right now a missing preimage and a broken blob both return `HUH`, so the caller can't tell "not available, maybe retry elsewhere" apart from "your code is garbage".

This returns `NONE` for the missing preimage case instead, same as `historical_lookup` does for the same failure.

davxy (dismissed):
LGTM. Just one suggestion

gavofyork:
...also needs resolving...

koute:
Merged `main` and resolved @davxy's comment.

davxy (approved):

gavofyork (approved):

PR #531, opened 2026-05-28
WiP PVM working-set

zdave-parity:
Incomplete draft of Jan's proposed PVM working-set mechanism. I've not yet updated the host calls section, working on that now. I think the other bits are pretty much complete.

One thing I'm not sure about is how to handle the working-set size for accumulation. I've currently got that in the service state (presumably would be updated by `upgrade`, this is not done yet). The reason for this is that it is needed for transfers as well as for accumulation of work-items. It would seem very awkward to require a working-set size to be specified in `transfer`, and if we did do something like that we would need some way to determine a single working-set size for an accumulate call which handles multiple transfers/work-items.


text/definitions.tex
```diff
@@ -294,6 +295,13 @@ \subsubsection{Constants}
   \item[$\Cmaxpackageexports = 3,072$] The maximum number of exports in a work-package. See equation \ref{eq:limitworkpackagebandwidth}.
   \item[$\mathsf{X}$] Context strings, see below.
   \item[$\Cepochtailstart = 500$] The number of slots into an epoch at which ticket-submission ends. See sections \ref{sec:slotkeysequence}, \ref{sec:epochmarker} and \ref{sec:safrolextandtickets}.
+  \item[$\Cpvmgaspagealloc = 300$] The gas cost of provisioning (and eventually deallocating) a physical frame for a \textsc{pvm} \textsc{ram} page. See appendix \ref{sec:workingsetchanges}.
+  \item[$\Cpvmgaspagecopy = 300$] The gas cost of populating a freshly provisioned \textsc{pvm} \textsc{ram} page from its initial image. See appendix \ref{sec:workingsetchanges}.
+  \item[$\Cpvmgasdeliver = 500$] The gas cost, per 1024 octets, of the eventual read-back, from the node's state or context, of data that a host-call writes into a machine's \textsc{ram}. See appendix \ref{sec:generalfunctions}.
```
zdave-parity, line 300: This description does not make sense to me... what does it mean for data that is written to machine RAM to be eventually read from the state/context?

text/pvm.tex
```diff
@@ -158,65 +176,89 @@ \subsection{Single-Step State Transition}
...
+\begin{equation}
+  \tup{\varepsilon^\phase¬fetch, \gascounter^\phase¬fetch, \workingset^\phase¬fetch} = \codepagecharge(\gascounter^\phase¬gas, \workingset^\phase¬gas, \pageforinstruction(\mathbf{c}, \mathbf{k}, \imath))
+\end{equation}
+
+An instruction that resolves its target through the jump-table reads an 8-octet entry from a native jump-table, laid out at a stride of eight octets per dynamic address unit; the entry for a target address $a$ thus lies on the jump-table page $\jumppage{\floor{\nicefrac{8a}{\Cpvmpagesize}}}$, where jump-table pages $\jumptag \times \N$ form their own class of page index, distinct from the \textsc{ram} and code pages of the same machine. This read is charged like any other access. No charge arises for the special halt address (its vector is permanently resident), nor where the entry's page lies wholly beyond the table (execution panics without any table access being made), however a charge \emph{is} made for a misaligned, zero or invalid-entry target whose page is within reach of the table, as the entry must be read before it can be found invalid. Formally:
```
zdave-parity, line 210: Using a `(tag, index)` tuple to identify a page in a machine makes sense but the syntax seems a bit confused here. `\times` is used for set product, so `A \times B` means `{ (a, b) | a \in A, b \in B }`. `\jumppage{x}` should be `(J, x)` and the set of jump-table pages should be written `\set{\jumptag} \times \N` rather than just `\jumptag \times \N`.

FWIW `\pageid` is now essentially `(machine, (tag, index))`. Might make sense to flatten that to `(machine, tag, index)`, would have to try it out to see if it made things simpler...

text/pvm.tex
```diff
@@ -158,65 +176,89 @@ \subsection{Single-Step State Transition}
...
-          \set{\kv{\top}{\readable\mem}, \kv{\bot}{\writable\mem}}\subb{r} \\
-        (\oog, g, w) &\otherwhen g < g^\delta \\
-        A(g - g^\delta, w', \mathbf{r}) &\otherwise
-      \end{cases} \\
-      &\where y = x \bmod 2^{32} \\
-      &\also \pid¬index = \floor{\frac{y}{\Cpvmpagesize}} \\
-      &\also (g^\delta, w') = \gascostforpageaccess(w, \tup{\pid¬machine, \pid¬index})
-    \end{aligned}
+    (\varepsilon^\phase¬mem, \gascounter^\phase¬mem, \workingset^\phase¬mem, \mem^\phase¬mem) &= A(\gascounter^\phase¬djump, \workingset^\phase¬djump, \mem,
+      \sqorderby{x}{\set{\build{(x, \is{r}{\top})}{x \in \mathbf{r}}} \cup \set{\build{(x, \is{r}{\bot})}{x \in \mathbf{w}}}}) \\
```
zdave-parity, line 230: Probably this should be something like:

    \sqorderby{x}{(x, r) \in \set{\build{(x, \is{r}{\top})}{x \in \mathbf{r}}} \cup \set{\build{(x, \is{r}{\bot})}{x \in \mathbf{w}}}})

To make it really clear where the `x` is coming from for ordering?

text/pvm_invocations.tex
```diff
@@ -819,9 +861,11 @@ \subsection{Accumulate Functions}\label{sec:accumulatefunctions}
         \sa¬codehash,
         \is{\mathbf{\sa¬storage}}{\emset},
         \is{\mathbf{\sa¬requests}}{\set{\kv{\tup{c, l}}{\sq{}}}},
+        % TODO: Should be configurable, but we're out of argument registers. Pass a pointer to struct instead?
```
zdave-parity, line 864: You mean the initial balance should be configurable? I did have that at one point in the authority services change (https://github.com/gavofyork/graypaper/pull/539) but ultimately dropped it as it will be possible to immediately transfer extra balance across after creation.

text/pvm_invocations.tex
```diff
@@ -291,6 +291,52 @@ \subsection{General Functions}\label{sec:generalfunctions}
...
+      P(g - g^\delta, \mem', \mathbf{i}\sub{1\dots}) &\otherwise
+    \end{cases} \\
+    \also g^\delta &= \pagefaultcost(\mem, \mathbf{i}_0) - \Cpvmgaspagefault \\
+    \also \mem' &= \mem \exc \mem'_\ram¬frames\subb{\mathbf{i}_0} = \frame¬present
+  \end{aligned}
+\end{equation}
+
+Such an access touches the octets of its range one at a time in ascending order of address, each octet an access of its containing page. Accessibility is checked before any charge, so these charges never fault, though they may exhaust gas.
+
+Should the gas run out mid-way, the result is $\oog$; the gas counter and the frames provisioned so far retain the charges already made, though no octets are written into \textsc{ram}. As $\oog$ discards the whole invocation's output, the surrounding context is immaterial and taken as its prior value. All of this is levied after the nominal $g$ of the tables below. In the descriptions that follow, the working-set argument and result are elided except where a function threads them explicitly.
```
zdave-parity, line 338: I think we can simplify this a bit as in the case of gas exhaustion we don't keep any state changes, we basically just return "OOG" from the invoke-host-call function. The gas costs for applying any working-set changes etc can simply be added to `g` along with the base & linear gas costs, letting the normal gas charging logic do its thing. That is, have `\wscharge¬extwrite` return eg `g_w` for the gas costs associated with the write and stick a `+ g_w` on the end of the `g` definition.

zdave-parity (commented):
Still looking through this. Your changes make sense to me so far, just a few minor comments. It would be nice to merge main, think there are likely to be conflicts with the host-call changes.

text/overview.tex
```diff
@@ -176,19 +177,26 @@ \subsection{The Virtual Machine and Gas}\label{sec:virtualmachineandgas}
...
   \ram &\equiv \tuple{
     \isa{\ram¬value}{\blob[2^{32}]},
-    \isa{\ram¬access}{\sequence[p]{\set{\text{W}, \text{R}, \none}}}
-  }\,,\ p = \frac{2^{32}}{\Cpvmpagesize}\\
+    \isa{\ram¬access}{\sequence[p]{\set{\text{W}, \text{R}, \none}}},
+    \isa{\ram¬frames}{\sequence[p]{\set{\frame¬present, \frame¬missing¬zero, \frame¬missing¬copy}}}
+  }\\
+  p &= \frac{2^{32}}{\Cpvmpagesize} \notag\\
   \Cpvmpagesize &= 2^{12}
 \end{align}
```
zdave-parity: Layout is a bit messy here. Maybe:

    \begin{gather}\label{eq:pvmmemory}
      \ram \equiv \tuple{
        \begin{aligned}
          &\isa{\ram¬value}{\blob[2^{32}]},
          \isa{\ram¬access}{\sequence[p]{\set{\text{W}, \text{R}, \none}}}, \\
          &\isa{\ram¬frames}{\sequence[p]{\set{\frame¬present, \frame¬missing¬zero, \frame¬missing¬copy}}}
        \end{aligned}
      }\,,\ p = \frac{2^{32}}{\Cpvmpagesize}\\
      \Cpvmpagesize = 2^{12}
    \end{gather}

text/overview.tex
```diff
@@ -176,19 +177,26 @@ \subsection{The Virtual Machine and Gas}\label{sec:virtualmachineandgas}
...
+The \textsc{pvm} assumes a simple pageable \textsc{ram} of 32-bit addressable octets situated in pages of $\Cpvmpagesize = 4096$ octets where each page may be either immutable, mutable or inaccessible. The \textsc{ram} definition $\ram$ includes three components: a value $\ram¬value$, access $\ram¬access$, and frames $\ram¬frames$. If the component is unspecified while being subscripted then the value component may be assumed. The frames component records, per page, whether a physical frame has been allocated for it ($\frame¬present$) by the working-set gas model of appendix \ref{sec:workingsetchanges}, or, if not, whether first provisioning one would require copying the page from the program's initial image ($\frame¬missing¬copy$) or merely zeroing it ($\frame¬missing¬zero$). It influences gas costs only and has no effect on the semantics of reads and writes. Within the context of the virtual machine, $\memory \in \ram$ is typically used to denote \textsc{ram}.
```
zdave-parity: Might be nice to move the bit about frames (starting at "The frames component records") to after the paragraph below where the working-set is discussed? Feels like things are introduced in the wrong order at the moment.

text/pvm.tex
```diff
@@ -25,25 +25,46 @@ \subsection{Basic Definition}
...
 \newcommand*{\smod}{\text{smod}}
 \newcommand*{\rtz}{\text{rtz}}
 \newcommand*{\gascostforblock}{\gascounter^{\Delta}}
+\newcommand*{\datatag}{\text{\textsc{d}}}
+\newcommand*{\codetag}{\text{\textsc{c}}}
+\newcommand*{\jumptag}{\text{\textsc{j}}}
+\newcommand*{\datapage}[1]{{\datatag \times #1}}
+\newcommand*{\codepage}[1]{{\codetag \times #1}}
+\newcommand*{\jumppage}[1]{{\jumptag \times #1}}
+\newcommand*{\codefam}[1]{{#1}^\diamond}
```
zdave-parity: Looks like you can probably drop the `^\diamond` stuff -- I don't think any of the unadorned names are used in this section?

text/pvm.tex
```diff
@@ -25,25 +25,46 @@ \subsection{Basic Definition}
...
+\newcommand*{\pageforinstruction}{\codefam{P}}
+\newcommand*{\preamblepages}{\codefam{Q}}
+\newcommand*{\codeoffset}{\codefam{o}}
```
zdave-parity: Would make this capital O -- generally all "local" functions use capital letters.

text/pvm.tex
```diff
@@ -25,25 +25,46 @@ \subsection{Basic Definition}
...
-      \Psi(\pvm¬blob, \imath', \gascounter', \gaschargedflag', \registers', \mem') &\when \varepsilon = \continue\\
-      \tup{\varepsilon, 0, \gascounter', \gaschargedflag', \registers', \mem'} &\when \varepsilon \in \set{ \panic, \halt }\\
-      \tup{\varepsilon, \imath, \gascounter', \gaschargedflag', \registers, \mem} &\otherwise
+    \tuple{\blob, \pvmreg, \gas, \bool, \regs, \ram, \optional{\N}, \sequence{\optional{\pageid}}} &\to \tuple{\set{\halt, \panic, \oog} \cup \set{\fault, \host} \times \pvmreg, \pvmreg, \gas, \bool, \regs, \ram, \sequence{\optional{\pageid}}}\\
+    \tup{\pvm¬blob, \imath, \gascounter, \gaschargedflag, \registers, \mem, \pid¬machine, \workingset} &\mapsto \begin{cases}
+      \Psi(\pvm¬blob, \imath', \gascounter', \gaschargedflag', \registers', \mem', \pid¬machine, \workingset') &\when \varepsilon = \continue\\
+      \tup{\varepsilon, 0, \gascounter', \gaschargedflag', \registers', \mem', \workingset'} &\when \varepsilon \in \set{ \panic, \halt }\\
+      \tup{\varepsilon, \imath, \gascounter', \gaschargedflag', \registers, \mem, \workingset'} &\otherwise
```
zdave-parity: I don't think this is quite right now for the case where we do a page-crossing access and the second page faults. We want to preserve the effect of any provisioning we did for the first page in `\mem`?

I think it would be clearest if we just had a single exceptional (ie non-continue) case here which returned all of the prime variables. That is, just defer to `Psi_1` for everything. It's a bit confusing that some of the return values of `Psi_1` are ignored in exceptional cases. It would probably make sense to rework the second paragraph below a bit if this change is made, as the relevant logic would live in `Psi_1` then...

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
   }
 \end{equation}
 
-On the very first step of execution, and every time the execution enters a new basic block or jumps back to the beginning of the current basic block, the gas counter of the machine is updated according to the gas cost function $\gascostforblock$~(\ref{eq:gascostforblock}) of the target basic block. No instruction is allowed to execute within a basic block unless the gas cost for the entire basic block has been charged in advance. In case there's not enough gas remaining to cover the full gas cost, the execution is interrupted and the gas counter remains unchanged. Formally:
+Before execution, the instruction must be fetched from memory. This is done lazily, using the machine's paging hardware: each fetch of a code page not resident in the working-set is a page fault which charges gas according to $\gascostforpageaccess$ and inserts the page. We define $\codepagecharge$, which charges in order for a sequence of pages, each inserted at the base fault cost $\Cpvmgaspagefault$, halting if the gas runs out:
```
zdave-parity: In what sense is this done lazily? I think just "This is done using the machine's paging hardware" is fine and less confusing.

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
+\begin{equation}
+  \begin{aligned}
+    \codepagecharge(g, w, \mathbf{p}) &= \begin{cases}
```
zdave-parity:
```suggestion
    \codepagecharge(g, \mathbf{w}, \mathbf{p}) &\equiv \begin{cases}
```

Generally `\mathbf` is used for composite types, so I would change `w` to `\mathbf{w}` here. Also `\equiv` is used for function definitions.

text/pvm.tex
zdave-parity:
```suggestion
Before execution, the instruction must be fetched from memory. This is done lazily, using the machine's paging hardware: each fetch of a code page not resident in the working-set is a page fault which charges gas according to $\gascostforpageaccess$ and inserts the page. We define $\codepagecharge$, which charges in order for a sequence of code pages, each inserted at the base fault cost $\Cpvmgaspagefault$, halting if the gas runs out:
```

To make it clear that `C` is only for code pages.

text/pvm.tex
zdave-parity: FWIW I've noticed that the `\times` notation is used for constructing page-fault and host-call returns, like `\fault \times i \cdot \Cpvmpagesize` below. This seems wrong to me, I would write eg `\tup{\fault, i \cdot \Cpvmpagesize}` instead. But maybe @gavofyork disagrees!

text/pvm_invocations.tex
koute: Sorry, I have no idea how I fat-fingered this. I meant to put this above the `\is{\sa¬accworkingsetsize}{\Cpvmmaxworkingset}` (i.e. make the working-set configurable).

text/pvm.tex
koute: Hmm... I don't think flattening will make it simpler, considering that'd either force a machine ID to be passed where it isn't needed *or* extra repacking churn, just to remove a single extra pair of parentheses?

text/definitions.tex
koute: I meant the cost of reading the data back by the VM -- i.e. host/node writes the data into the VM's memory, and because the VM is assumed to run in another L3 domain this invalidates its cache lines, so this is supposed to pay for when it tries to actually read that data and pull those cache lines back into its L3.

I'll try to make this more clear.

text/pvm.tex
```diff
@@ -25,25 +25,46 @@ \subsection{Basic Definition}
...
 \newcommand*{\rtz}{\text{rtz}}
 \newcommand*{\gascostforblock}{\gascounter^{\Delta}}
+\newcommand*{\datatag}{\text{\textsc{d}}}
+\newcommand*{\codetag}{\text{\textsc{c}}}
+\newcommand*{\jumptag}{\text{\textsc{j}}}
+\newcommand*{\datapage}[1]{{\tup{\datatag, #1}}}
+\newcommand*{\codepage}[1]{{\tup{\codetag, #1}}}
+\newcommand*{\jumppage}[1]{{\tup{\jumptag, #1}}}
+\newcommand*{\codefam}[1]{{#1}^\diamond}
+\newcommand*{\pageforinstruction}{\codefam{P}}
```
zdave-parity: I guess this should be `\pagesforinstruction` now, as it returns a sequence of pages?

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
+\end{equation}
 
-Formally, let $\mathbf{r}$ and $\mathbf{w}$ be the set of indices by which $\mem$ must be subscripted for inspection and mutation respectively in order to calculate the result of $\Psi_1$. We define the memory-access exceptional execution state $\varepsilon^\mu$ as following:
+An instruction that resolves its target through the jump-table reads an 8-octet entry from a native jump-table, laid out at a stride of eight octets per dynamic address unit; the entry for a target address $a$ thus lies on the jump-table page $\jumppage{\floor{\nicefrac{8a}{\Cpvmpagesize}}}$, where jump-table pages $\set{\jumptag} \times \N$ form their own class of page index, distinct from the \textsc{ram} and code pages of the same machine. This read is charged like any other access. No charge arises for the special halt address (its vector is permanently resident), nor where the entry's page lies wholly beyond the table (execution panics without any table access being made), however a charge \emph{is} made for a misaligned, zero or invalid-entry target whose page is within reach of the table, as the entry must be read before it can be found invalid. Formally:
```
zdave-parity:
```suggestion
An instruction that resolves its target through the jump-table reads an 8-octet entry from a native jump-table, laid out at a stride of eight octets per dynamic address unit; the entry for a target address $a$ thus lies on the jump-table page $\jumppage{\floor{\nicefrac{8a}{\Cpvmpagesize}}}$. Jump-table pages $\set{\jumptag} \times \N$ form their own class of page index, distinct from the \textsc{ram} and code pages of the same machine. This read is charged like any other access. No charge arises for the special halt address (its vector is permanently resident), nor where the entry's page lies wholly beyond the table (execution panics without any table access being made), however a charge \emph{is} made for a misaligned, zero or invalid-entry target whose page is within reach of the table, as the entry must be read before it can be found invalid. Formally:
```

Bit of a run-on sentence.

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
-    \end{cases}\\
-    \where \mathbf{x} &= \set{\build{x}{x \in \mathbf{r} \wedge x \bmod 2^{32} \not\in \readable\mem\ \vee\ x \in \mathbf{w} \wedge x \bmod 2^{32} \not\in \writable\mem}}
+    \tup{\varepsilon^\phase¬djump, \gascounter^\phase¬djump, \workingset^\phase¬djump} &= \codepagecharge(\gascounter^\phase¬fetch, \workingset^\phase¬fetch, \mathbf{p}^\phase¬djump) \\
+    \where \mathbf{p}^\phase¬djump &= \begin{cases}
+      \sq{} &\when a^\circ = \none \vee a^\circ = 2^{32} - 2^{16} \vee \floor{\nicefrac{8a^\circ}{\Cpvmpagesize}} > \floor{\nicefrac{8\Cpvmdynaddralign\len{\mathbf{j}}}{\Cpvmpagesize}} \\
+      \sq{\jumppage{\floor{\nicefrac{8a^\circ}{\Cpvmpagesize}}}} &\otherwise
+    \end{cases}
   \end{aligned}
 \end{equation}
+where $a^\circ \in \optional{\N}$ is the dynamic-jump target address which the instruction at $\imath$ resolves through the jump-table --- that is, the argument passed to its $\token{djump}$ (section \ref{sec:instructiontables}) --- or $\none$ where the instruction performs no dynamic jump.
```
zdave-parity:
```suggestion
Here, $a^\circ \in \optional{\N}$ is the dynamic-jump target address which the instruction at $\imath$ resolves through the jump-table --- that is, the argument passed to its $\token{djump}$ in section \ref{sec:instructiontables} --- or $\none$ where the instruction performs no dynamic jump.
```

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
   \begin{aligned}
-    \varepsilon^{\mu} &= \begin{cases}
-      \continue &\when \mathbf{x} = \emptyset \\
-      \panic &\when \min(\mathbf{x}) \bmod 2^{32} < 2^{16} \\
-      \fault \times \Cpvmpagesize\floor{\min(\mathbf{x}) \bmod 2^{32} \div \Cpvmpagesize} &\otherwise
-    \end{cases}\\
-    \where \mathbf{x} &= \set{\build{x}{x \in \mathbf{r} \wedge x \bmod 2^{32} \not\in \readable\mem\ \vee\ x \in \mathbf{w} \wedge x \bmod 2^{32} \not\in \writable\mem}}
+    \tup{\varepsilon^\phase¬djump, \gascounter^\phase¬djump, \workingset^\phase¬djump} &= \codepagecharge(\gascounter^\phase¬fetch, \workingset^\phase¬fetch, \mathbf{p}^\phase¬djump) \\
+    \where \mathbf{p}^\phase¬djump &= \begin{cases}
+      \sq{} &\when a^\circ = \none \vee a^\circ = 2^{32} - 2^{16} \vee \floor{\nicefrac{8a^\circ}{\Cpvmpagesize}} > \floor{\nicefrac{8\Cpvmdynaddralign\len{\mathbf{j}}}{\Cpvmpagesize}} \\
```
zdave-parity: You also want an `a^\circ = 0` case here to be consistent with the prose?

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
-\begin{align}
-  \tup{\varepsilon^*, \imath^*, \registers^*, \mem^*} &= \begin{cases}
-    \tup{\varepsilon^{\gascounter}, \imath, \registers, \mem} &\when \varepsilon^{\gascounter} \neq \continue \\
-    \tup{\varepsilon^{\mu}, \imath, \registers, \mem} &\otherwhen \varepsilon^{\mu} \neq \continue \\
-    \tup{\varepsilon, \imath', \registers', \mem'} &\otherwise
-  \end{cases}
-\end{align}
+During the course of executing instructions \textsc{ram} may be accessed. When an index of \textsc{ram} below $2^{16}$ is required, the machine always panics immediately without further changes to its state regardless of the apparent (in)accessibility of the value. Otherwise, should the given index of \textsc{ram} not be accessible then machine state remains unchanged and the exit reason is a fault with the lowest inaccessible \emph{page address} to be read. Similarly, where \textsc{ram} must be mutated and yet mutable access is not possible, then machine state is unchanged, and the exit reason is a fault with the lowest page address to be written which is inaccessible.
```
zdave-parity:
```suggestion
During the course of executing an instruction \textsc{ram} may be accessed. When an index of \textsc{ram} below $2^{16}$ is required, the machine always panics immediately regardless of the apparent (in)accessibility of the value. Otherwise, should the given index of \textsc{ram} not be accessible then the exit reason is a fault with the lowest inaccessible \emph{page address} to be read. Similarly, where \textsc{ram} must be mutated and yet mutable access is not possible, then the exit reason is a fault with the lowest page address to be written which is inaccessible. In all of these exceptional cases, the program counter, registers, and memory remain unchanged.
```

I think particularly now that we have the working-set we should be more careful about saying that "machine state is unchanged", or alternatively we can clarify somewhere that when we say this we only mean eg the PC, registers, and memory.

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
 
-We also have to adjust the state based on the executed instruction, to force a gas charge on the next step when necessary:
+In addition to these accessibility checks, as with instruction fetches, \textsc{ram} reads and writes implicitly pull accessed pages into the working-set, allocating a frame for the page (recorded in $\mem_\ram¬frames$) where none yet exists. If the gas cost of a working-set change cannot be covered by the available gas, execution is interrupted. Charges made (and frames allocated) for the earlier pages of a step persist even where a later access of the same step faults or exhausts gas.
```
zdave-parity:
```suggestion
In addition to these accessibility checks, as with instruction fetches, \textsc{ram} reads and writes implicitly pull accessed pages into the working-set, allocating frames for pages (recorded in $\mem_\ram¬frames$) where none yet exist. If the gas cost of a working-set change cannot be covered by the available gas, execution is interrupted. Charges made (and frames allocated) for the earlier pages of a step persist even where a later access of the same step faults or exhausts gas.
```

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
+
+Formally, let $\mathbf{r}$ and $\mathbf{w}$ be the set of indices by which $\mem$ must be subscripted for inspection and mutation respectively in order to calculate the result of $\Psi_1$. We define the memory-access exceptional execution state $\varepsilon^\phase¬mem$, posterior gas balance $\gascounter^\phase¬mem$, posterior working-set $\workingset^\phase¬mem$, and posterior \textsc{ram} $\mem^\phase¬mem$ (which may differ from $\mem$ in its frames component only) as following:
```
zdave-parity:
```suggestion
Formally, let $\mathbf{r}$ and $\mathbf{w}$ be the set of indices by which $\mem$ must be subscripted for inspection and mutation respectively in order to calculate the result of $\Psi_1$. We define the memory-access exceptional execution state $\varepsilon^\phase¬mem$, posterior gas balance $\gascounter^\phase¬mem$, posterior working-set $\workingset^\phase¬mem$, and posterior \textsc{ram} $\mem^\phase¬mem$ (which may differ from $\mem$ in its frames component only) as follows:
```

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
 \begin{equation}
-  \gaschargedflag^* = \begin{cases}
-    \bot &\when \gaschargedflag' = \bot \lor (\mathbf{c}_\imath \in T \land \varepsilon^* \in \set{\continue} \cup \set{\host} \times \pvmreg) \\
-    \top &\otherwise
+  \begin{aligned}
+    (\varepsilon^\phase¬mem, \gascounter^\phase¬mem, \workingset^\phase¬mem, \mem^\phase¬mem) &= A(\gascounter^\phase¬djump, \workingset^\phase¬djump, \mem,
+      \sqorderby{x}{(x, r) \in \set{\build{(x, \is{r}{\top})}{x \in \mathbf{r}}} \cup \set{\build{(x, \is{r}{\bot})}{x \in \mathbf{w}}}}) \\
+    A(g, w, \mem, \mathbf{p}) &= \begin{cases}
```
zdave-parity: As above should probably use `\mathbf{w}` for the working-set as it has a composite type.

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
+      (\panic, g, w, \mem) &\otherwhen y < 2^{16} \\
+      (\fault \times i \cdot \Cpvmpagesize, g, w, \mem) &\otherwhen y \not\in
+        \set{\kv{\top}{\readable\mem}, \kv{\bot}{\writable\mem}}\subb{r} \\
+      (\oog, g, w, \mem) &\otherwhen g < g^\delta \\
+      A(g - g^\delta, w', \mem', \mathbf{p}\sub{1\dots}) &\otherwise
+    \end{cases} \\
+    \where \tup{x, r} &= \mathbf{p}_0 \\
+    \also y &= x \bmod 2^{32} \\
+    \also i &= \floor{\frac{y}{\Cpvmpagesize}} \\
+    \also (g^\delta, w') &= \gascostforpageaccess(w, \tup{\pid¬machine, \datapage{i}}, \pagefaultcost(\mem, i)) \\
```
zdave-parity:
```suggestion
    \also (g^\delta, w') &= \gascostforpageaccess(w, \tup{\pid¬machine, \is{\pid¬index}{\datapage{i}}}, \pagefaultcost(\mem, i)) \\
```

text/pvm.tex
```diff
@@ -150,55 +171,94 @@ \subsection{Single-Step State Transition}
...
+    \also \mem' &= \mem \exc \mem'_\ram¬frames\subb{i} = \frame¬present
+  \end{aligned}
+\end{equation}
+
```
zdave-parity: We should stick a pointer to the `\pagefaultcost` definition in the text and maybe give a short description of what it does. Perhaps something like:

    $\pagefaultcost(\mem, i)$ is the cost of inserting page $i$ into the working-set, including the cost of provisioning and initialising the page as necessary. $\pagefaultcost$ is defined in equation XXX.

text/pvm.tex
```diff
@@ -752,35 +812,39 @@ \subsubsection{Instructions with Arguments of Three Registers}
...
-\end{align}
+    \begin{aligned}
+      \Psi_H(\pvm¬blob, \imath, \gascounter, \registers, \mem, w, f, \context) &\equiv (\varepsilon, \imath', \gascounter', \registers', \mem', \context') \\
+      \where (\varepsilon, \imath', \gascounter', \gaschargedflag', \registers', \mem', \workingset', \context') &=
+        \Psi_H^*(\pvm¬blob, \imath, \gascounter, \bot, \registers, \mem, \sq{\none, \none, \dots}\sub{\dots w}, f, \context)
+    \end{aligned}\\
+    \contextmutator{X} \equiv \tuple{\N, \gas, \regs, \ram, \sequence{\optional{\pageid}}, X} \to \tuple{\set{\continue, \panic, \oog}, \gas, \regs, \ram, \sequence{\optional{\pageid}}, X}
+\end{gather}
 
 As with $\Psi$, on exit the instruction counter references the instruction \emph{which caused the exit} and the machine state is that prior to this instruction. Should the machine be invoked again using this instruction counter and code, then the same instruction which caused the exit would be executed on the proper (prior) machine state.
```
zdave-parity:
```suggestion
As with $\Psi$, on exit the instruction counter references the instruction \emph{which caused the exit} and the register/memory contents and context are those prior to this instruction. Should the machine be invoked again using this instruction counter and code, then the same instruction which caused the exit would be executed on the proper (prior) machine state.
```

zdave-parity: "the machine state is that prior to this instruction" is not really true any more given that we can eg provision pages before returning an error. We could say "the memory/register contents and context are those prior to this instruction", though actually that depends on the definition of `f` and so can't strictly speaking be said in general (it is true for all of the context mutator functions we define though).

text/pvm.tex
```diff
@@ -850,17 +914,25 @@ \subsection{Standard Program Initialization}\label{sec:standardprograminit}
     \end{cases}
 \end{equation}
 
+No page begins with an allocated frame. The cost of provisioning each page's frame falls on whichever access first touches it (see appendix \ref{sec:workingsetchanges}), and depends on whether that page's initial content must be copied from the program image or is merely zeroed:
+\begin{equation}
+  \forall j \in \Nmax{\nicefrac{2^{32}}{\Cpvmpagesize}} : \mem_\ram¬frames\subb{j} = \begin{cases}
+    \frame¬missing¬copy &\when \Nrange{j\Cpvmpagesize}{\Cpvmpagesize} \cap \mathcal{I} \ne \emptyset \\
```
zdave-parity:
```suggestion
    \frame¬missing¬copy &\when \Nrange{j\Cpvmpagesize}{\Cpvmpagesize} \cap \mathcal{I} \ne \emset \\
```

I think generally `\emset` (= {}) is used for the empty set rather than `\emptyset`.

text/pvm.tex
```diff
@@ -850,17 +914,25 @@ \subsection{Standard Program Initialization}\label{sec:standardprograminit}
     \end{cases}
 \end{equation}
 
+No page begins with an allocated frame. The cost of provisioning each page's frame falls on whichever access first touches it (see appendix \ref{sec:workingsetchanges}), and depends on whether that page's initial content must be copied from the program image or is merely zeroed:
+\begin{equation}
```
zdave-parity: There is space to put this in the equations above if we put the `\forall i \in \Nbits{32}` on its own line? That would be quite a bit clearer I think.

Something like this above:

    \begin{equation}\label{eq:memlayout}
      \begin{aligned}
        \forall i \in \Nbits{32} &: \\
        ((\mem_\ram¬value)\sub{i}, (\mem_\ram¬access)_{\floor{\nicefrac{i}{\Cpvmpagesize}}}, (\mem_\ram¬frames)_{\floor{\nicefrac{i}{\Cpvmpagesize}}}) &= \bracegroup{\begin{alignedat}{5}
          &\tup{\is{\ram¬value}{\mathbf{o}_{i - \Cpvminitzonesize}},\,\is{\ram¬access}{R}, \frame¬missing¬copy} &&\ \when
              \Cpvminitzonesize
                  &\ \leq i < \ &&
                      \Cpvminitzonesize + \len{\mathbf{o}}\\
          &\tup{0, R, \frame¬missing¬zero} &&\ \when
              \Cpvminitzonesize + \len{\mathbf{o}}
                  &\ \leq i < \ &&
                      \Cpvminitzonesize + \rnp{\len{\mathbf{o}}} \\
          &(\mathbf{w}_{i - (2\Cpvminitzonesize + \rnq{\len{\mathbf{o}}})}, W, \frame¬missing¬copy) &&\ \when
              2\Cpvminitzonesize + \rnq{\len{\mathbf{o}}}
                  &\ \leq i < \ &&
                      2\Cpvminitzonesize + \rnq{\len{\mathbf{o}}} + \len{\mathbf{w}}\\
          &\tup{0, W, \frame¬missing¬zero} &&\ \when
              2\Cpvminitzonesize + \rnq{\len{\mathbf{o}}} + \len{\mathbf{w}}
                  &\ \leq i < \ &&
                      2\Cpvminitzonesize + \rnq{\len{\mathbf{o}}} + \rnp{\len{\mathbf{w}}} + z\Cpvmpagesize\\
          &\tup{0, W, \frame¬missing¬zero} &&\ \when
              2^{32} - 2\Cpvminitzonesize - \Cpvminitinputsize - \rnp{s}
                  &\ \leq i < \ &&
                      2^{32} - 2\Cpvminitzonesize - \Cpvminitinputsize\\
          &(\mathbf{a}_{i - (2^{32} - \Cpvminitzonesize - \Cpvminitinputsize)}, R, \frame¬missing¬copy) &&\ \when
              2^{32} - \Cpvminitzonesize - \Cpvminitinputsize
                  &\ \leq i < \ &&
                      2^{32} - \Cpvminitzonesize - \Cpvminitinputsize + \len{\mathbf{a}}\\
          &\tup{0, R, \frame¬missing¬zero} &&\ \when
              2^{32} - \Cpvminitzonesize - \Cpvminitinputsize + \len{\mathbf{a}}
                  &\ \leq i < \ &&
                      2^{32} - \Cpvminitzonesize - \Cpvminitinputsize + \rnp{\len{\mathbf{a}}}\\
          &\tup{0, \none, \frame¬missing¬zero} &&\otherwise&&&
        \end{alignedat}}\\
      \end{aligned}
    \end{equation}

And here we can just say eg:

    Note that no page begins with an allocated frame; the cost of provisioning each page's frame falls on whichever access first touches it (see section \ref{sec:workingsetchanges}), and depends on whether that page's initial content must be copied from the program image ($\frame¬missing¬copy$) or is merely zeroed ($\frame¬missing¬zero$).

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
   }
 \end{equation}
 
+\subsection{Working-Set Changes}\label{sec:workingsetchanges}
+
+In addition to the static gas cost of each basic block as calculated by $\gascostforblock$, gas is charged dynamically whenever a memory page outside of the current working-set is accessed. Such an access implicitly adds the accessed page to the working-set, pushing another page out if necessary. The working-set is a first-in first-out buffer of fixed length, and each insertion drops the oldest entry. A dropped page ceases to be resident, and re-accessing it later incurs a further working-set charge. This affects only gas, never the contents or accessibility of the page, so (given enough gas) the working-set has no effect on execution semantics. The function $\gascostforpageaccess$ determines the working-set change, if any, required to access a page $p \in \pageid$ whose insertion cost is $c$, and the corresponding gas cost.
```
zdave-parity:
```suggestion
In addition to the static gas cost of each basic block as calculated by $\gascostforblock$, gas is charged dynamically whenever a memory page outside of the current working-set is accessed. Such an access implicitly adds the accessed page to the working-set. The working-set is a first-in first-out buffer of fixed length, and each insertion drops the oldest entry. A dropped page ceases to be resident, and re-accessing it later incurs a further working-set charge. This affects only gas, never the contents or accessibility of the page, so (given enough gas) the working-set has no effect on execution semantics. The function $\gascostforpageaccess$ determines the working-set change, if any, required to access a page $p \in \pageid$ whose insertion cost is $c$, and the corresponding gas cost.
```

A bit redundant.

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+\begin{equation}
+  \gascostforpageaccess\colon \abracegroup{
+    \tuple{\sequence{\optional{\pageid}}, \pageid, \N} &\to \tuple{\gas, \sequence{\optional{\pageid}}} \\
```
zdave-parity:
```suggestion
    \tuple{\sequence{\optional{\pageid}}, \pageid, \gas} &\to \tuple{\gas, \sequence{\optional{\pageid}}} \\
```

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+\begin{equation}\label{eq:pagefaultcost}
+  \pagefaultcost\colon \abracegroup{
+    \tuple{\ram, \N} &\to \N \\
```
zdave-parity:
```suggestion
    \tuple{\ram, \N} &\to \gas \\
```

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
...work implied by insertion and eviction. The insertion cost of \textsc{ram} pages additionally depends on whether a physical frame has previously been allocated for them, as recorded by the frames component $\ram¬frames$ of the \textsc{ram} (\ref{eq:pvmmemory}). Upon a page's first-ever access a frame must be provisioned, paying $\Cpvmgaspagealloc$ (which accounts both for its allocation and for its eventual deallocation when the machine is destroyed), as well as the cost of populating the page: $\Cpvmgaspagezero$ where the page is zero-sourced (the frame has to be merely cleared), and $\Cpvmgaspagecopy$ where it is copy-sourced (its content must be copied from the program's initial image). Which applies is a fixed property of the page's provenance, recorded as $\frame¬missing¬zero$ or $\frame¬missing¬copy$ in $\ram¬frames$ when the machine is laid out (\ref{eq:memlayout}), and does \emph{not} depend on the page's current contents. Once allocated, a frame persists regardless of eviction from the working-set, and re-insertion pays only the base cost. Formally, the insertion cost of the \textsc{ram} page with index $i$ of a machine with \textsc{ram} $\mu$ is given by $\pagefaultcost$:
```
zdave-parity:
```suggestion
The insertion cost of a code or a jump-table page is always the base page-fault cost $\Cpvmgaspagefault$, covering the fault itself together with the mapping and unmapping work implied by insertion and eviction. The insertion cost of a \textsc{ram} page additionally depends on whether a physical frame has previously been allocated for it, as recorded by the frames component $\ram¬frames$ of the \textsc{ram} (\ref{eq:pvmmemory}). Upon a page's first-ever access a frame must be provisioned, paying $\Cpvmgaspagealloc$ (which accounts both for its allocation and for its eventual deallocation when the machine is destroyed), as well as the cost of populating the page: $\Cpvmgaspagezero$ where the page is zero-sourced (the frame has to be merely cleared), and $\Cpvmgaspagecopy$ where it is copy-sourced (its content must be copied from the program's initial image). Which applies is a fixed property of the page's provenance, recorded as $\frame¬missing¬zero$ or $\frame¬missing¬copy$ in $\ram¬frames$ when the machine is laid out (\ref{eq:memlayout}), and does \emph{not} depend on the page's current contents. Once allocated, a frame persists regardless of eviction from the working-set, and re-insertion pays only the base cost. Formally, the insertion cost of the \textsc{ram} page with index $i$ of a machine with \textsc{ram} $\mu$ is given by $\pagefaultcost$:
```

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+\begin{equation}\label{eq:pagefaultcost}
+  \pagefaultcost\colon \abracegroup{
+    \tuple{\ram, \N} &\to \N \\
+    \tup{\mu, i} &\mapsto \begin{cases}
+      \Cpvmgaspagefault &\when \mu_\ram¬frames\subb{i} = \frame¬present \\
+      \Cpvmgaspagefault + \Cpvmgaspagealloc + \Cpvmgaspagezero &\otherwhen \mu_\ram¬frames\subb{i} = \frame¬missing¬zero \\
+      \Cpvmgaspagefault + \Cpvmgaspagealloc + \Cpvmgaspagecopy &\otherwise
```
zdave-parity:
```suggestion
      \Cpvmgaspagefault + \Cpvmgaspagealloc + \Cpvmgaspagecopy &\when \mu_\ram¬frames\subb{i} = \frame¬missing¬copy
```

Nit... maybe clearer to write all the conditions out with no "otherwise"s.

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+\begin{equation}
+  \gascostforpageaccess\colon \abracegroup{
+    \tuple{\sequence{\optional{\pageid}}, \pageid, \N} &\to \tuple{\gas, \sequence{\optional{\pageid}}} \\
+    \tup{\workingset, p, c} &\mapsto \begin{cases}
+      (0, \workingset) &\when p \in \workingset \\
+      (c, \workingset\sub{1\dots} \append p) &\otherwise
+    \end{cases}
+  }
+\end{equation}
```
zdave-parity: I think we should put the following here:

    This mechanism applies to explicit data accesses (via load/store instructions), as well as to the implicit reading of instructions by the machine. The instructions for a machine are not present in its \textsc{ram} $\memory$, and so instruction addresses cannot be mapped to regular page indices; instead they are mapped to special code-page indices $\set{\codetag} \times \N$. Similarly, jump-table pages have special indices $\set{\jumptag} \times \N$.

And remove this stuff from the paragraph below; it currently seems rather out of place.

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+    \tup{\mu, i} &\mapsto \begin{cases}
+      \Cpvmgaspagefault &\when \mu_\ram¬frames\subb{i} = \frame¬present \\
+      \Cpvmgaspagefault + \Cpvmgaspagealloc + \Cpvmgaspagezero &\otherwhen \mu_\ram¬frames\subb{i} = \frame¬missing¬zero \\
+      \Cpvmgaspagefault + \Cpvmgaspagealloc + \Cpvmgaspagecopy &\otherwise
+    \end{cases}
+  }
+\end{equation}
+
+A successful charge for a \textsc{ram} page always sets the page's entry in $\ram¬frames$ to $\frame¬present$. In case the charge cannot be met, machine state is left unchanged. This mechanism applies to explicit data accesses (via load/store instructions), as well as to the implicit reading of instructions by the machine. The instructions for a machine are not present in its \textsc{ram} $\memory$, and so instruction addresses cannot be mapped to regular page indices; instead they are mapped to special code-page indices $\set{\codetag} \times \N$. Assuming the native code is laid out contiguously from offset zero, we first define $\bbpreamblesize(\imath)$, the size of the gas-metering stub emitted before an instruction which begins a basic block (to charge that block's gas according to $\gascostforblock$):
```
zdave-parity:
```suggestion
A successful charge for a \textsc{ram} page always sets the page's entry in $\ram¬frames$ to $\frame¬present$. In case the charge cannot be met, machine state is left unchanged.

The gas model assumes \textsc{pvm} instructions are compiled to native instructions prior to execution, with the inclusion of a gas-metering preamble at the beginning of each basic block. Code page indices are derived from conservative estimates of the native code size required for each instruction and basic-block preamble. It is expected that no-op instructions are inserted into the native code as necessary to make the actual layout match the estimation. We first define $\bbpreamblesize(\imath)$, the size of the gas-metering stub emitted before the instruction at address $\imath$ (to charge that block's gas according to $\gascostforblock$), or 0 if the instruction does not begin a basic block:
```

I think it would be good here to quickly explain that code pages are derived from native code size estimates, something like this?

As commented above, I think the "This mechanism applies..." stuff should go earlier.

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+  \spannedpages(o, z) &\equiv\label{eq:pagespan} \begin{cases}
+    \sq{} &\when z = 0 \\
+    \sqorderby{p}{p \in \set{\build{p}{\floor{\frac{o}{\Cpvmpagesize}} \le p \le \floor{\frac{o + z - 1}{\Cpvmpagesize}}}}} &\otherwise
+  \end{cases} \\
+  \codepages(o, z) &\equiv \sq{\build{\codepage{p}}{p \orderedin \spannedpages(o, z)}} \\
+  \pageforinstruction(\mathbf{c}, \mathbf{k}, \imath) &\equiv \codepages(\codeoffset(\mathbf{c}, \mathbf{k}, \imath) + \bbpreamblesize(\imath),\ \inst¬compiledsize(\mathbf{c}, \mathbf{k}, \imath)) \\
+  \preamblepages(\mathbf{c}, \mathbf{k}, \imath) &\equiv \codepages(\codeoffset(\mathbf{c}, \mathbf{k}, \imath),\ \bbpreamblesize(\imath))
+\end{align}
+
+These are defined under the assumption that \textsc{pvm} program code is translated into native code for execution. The function $\inst¬compiledsize(\mathbf{c}, \mathbf{k}, \imath)$ (defined in \ref{sec:compiledsizes}) returns the size of the native code translation of the \textsc{pvm} instruction at address $\imath$.
```
zdave-parity: I would move this paragraph after the definition of `\bbpreamblesize` (and adjust the wording appropriately)?

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+      \inst¬compiledsize(\mathbf{c}, \mathbf{k}, p) + \bbpreamblesize(p) + L(\mathbf{c}, \mathbf{k}, \imath, p + 1 + \Fskip(p)) &\otherwise
+    \end{cases}
+  }
+\end{equation}
+
+The instruction's own native code occupies the byte range after its block's preamble (if any), and the preamble, where present, the range before it. $\spannedpages(o, z)$ gives the ascending pages spanned by the $z$-byte range at $o$, $\codepages$ tags them as code pages, and $\pageforinstruction$ and $\preamblepages$ apply this to the instruction's own range and to its preamble's:
+\begin{align}
+  \spannedpages(o, z) &\equiv\label{eq:pagespan} \begin{cases}
+    \sq{} &\when z = 0 \\
+    \sqorderby{p}{p \in \set{\build{p}{\floor{\frac{o}{\Cpvmpagesize}} \le p \le \floor{\frac{o + z - 1}{\Cpvmpagesize}}}}} &\otherwise
```
zdave-parity:
```suggestion
    \sq{\build{p}{p \orderedin \N, \floor{\frac{o}{\Cpvmpagesize}} \le p \le \floor{\frac{o + z - 1}{\Cpvmpagesize}}}} &\otherwise
```

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+  \pageforinstruction(\mathbf{c}, \mathbf{k}, \imath) &\equiv \codepages(\codeoffset(\mathbf{c}, \mathbf{k}, \imath) + \bbpreamblesize(\imath),\ \inst¬compiledsize(\mathbf{c}, \mathbf{k}, \imath)) \\
+  \preamblepages(\mathbf{c}, \mathbf{k}, \imath) &\equiv \codepages(\codeoffset(\mathbf{c}, \mathbf{k}, \imath),\ \bbpreamblesize(\imath))
+\end{align}
+
+These are defined under the assumption that \textsc{pvm} program code is translated into native code for execution. The function $\inst¬compiledsize(\mathbf{c}, \mathbf{k}, \imath)$ (defined in \ref{sec:compiledsizes}) returns the size of the native code translation of the \textsc{pvm} instruction at address $\imath$.
+
+The working-set is shared with inner \textsc{pvm} instances. For this reason, page identifiers characterized by the set $\pageid$ consist of a machine identifier $\pid¬machine$ in addition to a page index $\pid¬index$, the latter a page number paired with one of the disjoint class tags $\datatag$, $\codetag$ or $\jumptag$, for \textsc{ram}, code and jump-table pages respectively. $\none$ is used to identify the top-level \textsc{pvm} instance.
```
zdave-parity:
```suggestion
The working-set is shared with inner \textsc{pvm} instances. For this reason, page identifiers characterized by the set $\pageid$ consist of a machine identifier $\pid¬machine$ in addition to a page index $\pid¬index$. $\none$ is used to identify the top-level \textsc{pvm} instance. The page index is a page number paired with one of the disjoint class tags $\datatag$, $\codetag$ or $\jumptag$, for \textsc{ram}, code and jump-table pages respectively.
```

text/pvm.tex
```diff
@@ -1131,6 +1204,80 @@ \subsection{Gas Cost Model}\label{sec:gascostmodel}
...
+\begin{equation}
+  \label{eq:pageid}
+  \pageid \equiv \tuple{\isa{\pid¬machine}{\optional{\N}}, \isa{\pid¬index}{\set{\datatag, \codetag, \jumptag} \times \N}}
+\end{equation}
+
+The working-set size is fixed for the duration of an invocation, drawn from the valid sizes $\workingsetsize$~(\ref{eq:workingsetsize}). The lower bound $\Cpvmminworkingset$ guarantees forward progress: it exceeds the most pages any single step can need resident at once, so a step is never forced to evict a page it still requires. The upper bound $\Cpvmmaxworkingset$ keeps the resident set within the last-level cache of reference hardware, which is what justifies the pricing of resident accesses set out later in the gas cost tables~(\ref{eq:loadcost}).
+
+An entry is removed when the page it identifies ceases to exist: on expunging an inner machine, or deallocating or resetting \textsc{ram} pages through the $\mathtt{pages}$ host-call (appendix \ref{sec:refinefunctions}). The removed pages are replaced by $\none$ in-place. We write the removal of a page set $\mathbf{s} \subseteq \pageid$ as set-subtraction:
```
zdave-parity: I would write this as a function rather than overloading `\setminus`? As the natural definition of `\setminus` would have the elements removed rather than replaced with `\none`.

text/pvm.tex
zdave-parity:
```suggestion
    A(g, w, \mem, \mathbf{p}) &\equiv \begin{cases}
```

text/pvm.tex
zdave-parity:
```suggestion
An entry is removed when the page it identifies ceases to exist, \ie on expunging the page's machine, or on deallocating or resetting the page through the $\mathtt{pages}$ host-call (appendix \ref{sec:refinefunctions}). Removed pages are replaced by $\none$ in-place. We write the removal of a page set $\mathbf{s} \subseteq \pageid$ as set-subtraction:
```

text/pvm.tex
```diff
@@ -1160,13 +1307,17 @@ \subsection{Gas Cost Tables}\label{sec:gascosttables}
...
-The cost of memory accesses is defined as follows:
+The cost of memory loads depends on the size of the working-set $w$: the resident set always fits within the last-level cache, but a larger one exceeds the reach of the address-translation caches, making the worst-case resident access dearer. It is defined as follows:\footnote{These values are provisional, pending calibration; see appendix \ref{sec:workingsetchanges}.}
 
-\begin{equation}
-  \mathfrak{m} \equiv 25
+\begin{equation}\label{eq:loadcost}
+  \mathfrak{m}\colon \abracegroup{
+      \workingsetsize &\to \N \\
+      w &\mapsto \begin{cases}
+        25 &\when w \le 2^9 \\
```
zdave-parity:
```suggestion
        25 &\otherwhen w \le 2^9 \\
```

text/pvm.tex
```diff
@@ -1393,3 +1545,238 @@ \subsection{Gas Cost Tables}\label{sec:gascosttables}
...
...ompiledsize$ is given by the following table. $\srcdsteq{i}$ holds when the destination register coincides with the $i$-th source register; $\reg_b$ is the base register used by a load or store for address computation; and $t_1 = 3$ and $t_2 = 4$ are the two guest registers (\textsc{pvm}'s $\mathtt{T1}$ and $\mathtt{T2}$) whose underlying native registers force, respectively, a displacement byte and an index byte in native addressing. The decoded immediates $\immed_X, \immed_Y$ are as defined in section \ref{sec:instructiontables}. The direct branch and jump instructions are an exception: each consists of a fixed prefix (a register comparison or an immediate load) followed by a relative jump to its target basic block, and the translator encodes that jump in a compact 2-byte form precisely when the target has \emph{already} been translated (it lies at or before $\imath$, a backward branch) and is within a signed byte's reach; otherwise a longer form is used. Their size is therefore layout-dependent, given by $\branchsize(t, p, r, R)$ for a branch whose target is basic block $t$, whose fixed prefix is $p$ bytes, and whose trailing jump occupies $r$ bytes when short and $R$ when long:
```
zdave-parity: "T1" and "T2" are not mentioned anywhere else in the GP AFAICT so not sure it's useful to say "PVM's T1 and T2" here? Unless I'm missing something.

zdave-parity:
```suggestion
$\inst¬compiledsize$ is given by the table below. $\srcdsteq{i}$ holds when the destination register coincides with the $i$-th source register; $\reg_b$ is the base register used by a load or store for address computation; and $t_1 = 3$ and $t_2 = 4$ are the two guest registers whose underlying native registers are expected to force, respectively, a displacement byte and an index byte in native addressing. The decoded immediates $\immed_X, \immed_Y$ are as defined in section \ref{sec:instructiontables}.

Direct branch and jump instructions are expected to be translated to a fixed prefix (a register comparison or an immediate load) followed by a relative jump to the target basic block. It is expected that the translator encodes the jump in a compact 2-byte form precisely when the target has already been translated, \ie it lies at or before $\imath$ (a backward branch) and is within a signed byte's reach; otherwise a longer form is used. The size of direct branch and jump instructions is therefore layout-dependent, given by $\branchsize(t, p, r, R)$ for a branch whose target is basic block $t$, whose fixed prefix is $p$ bytes, and whose trailing jump occupies $r$ bytes when short and $R$ when long:
```

text/pvm.tex
```diff
@@ -1393,3 +1545,238 @@ \subsection{Gas Cost Tables}\label{sec:gascosttables}
...
+\begin{equation}
+  \branchsize(t, p, r, R) \equiv p + \begin{cases}
+    r &\when t \le \imath \;\wedge\; -128 \le \Delta \le 127 \\
+    R &\otherwise
+  \end{cases}
+  \quad\where \Delta = \codeoffset(\mathbf{c}, \mathbf{k}, t) - \left(\codeoffset(\mathbf{c}, \mathbf{k}, \imath) + \bbpreamblesize(\imath) + p + r\right)
+\end{equation}
+For the register-immediate branches the comparison prefix $q$ is smaller when the immediate fits in a signed byte:
+\begin{equation}
+  q = \begin{cases} 4 &\when -128 \le \signed{\immed_X} \le 127 \\ 7 &\otherwise \end{cases}
```
zdave-parity:
```suggestion
  q = \begin{cases} 4 &\when {-128} \le \signed{\immed_X} \le 127 \\ 7 &\otherwise \end{cases}
```

(To fix spacing)

text/pvm.tex
```diff
@@ -1393,3 +1545,238 @@ \subsection{Gas Cost Tables}\label{sec:gascosttables}
...
+    #1 &\when \srcdsteq{1} \\
+    #2 &\otherwise
+  \end{cases}$}
+\newcommand*{\srctwo}[2]{$\begin{cases}
+    #1 &\when \srcdsteq{1} \lor \srcdsteq{2} \\
+    #2 &\otherwise
+  \end{cases}$}
+\newcommand*{\srcimm}[3]{$\begin{cases}
+    #1 &\when \srcdsteq{1} \land {-}128 \le \signed{\immed_X} \le 127 \\
+    #2 &\when \srcdsteq{1} \lor {-}128 \le \signed{\immed_X} \le 127 \\
```
zdave-parity:
```suggestion
    #2 &\otherwhen \srcdsteq{1} \lor {-}128 \le \signed{\immed_X} \le 127 \\
```

And in the other commands...

text/pvm.tex
```diff
@@ -1393,3 +1545,238 @@ \subsection{Gas Cost Tables}\label{sec:gascosttables}
...
+    #3 &\otherwise
+  \end{cases}$}
+\newcommand*{\setltimm}{$\begin{cases}
+    12 &\when \srcdsteq{1} \land {-}128 \le \signed{\immed_X} \le 127 \\
+    15 &\when \srcdsteq{1} \\
+    11 &\when \lnot \srcdsteq{1} \land {-}128 \le \signed{\immed_X} \le 127 \\
```
zdave-parity:
```suggestion
    11 &\otherwhen {-}128 \le \signed{\immed_X} \le 127 \\
```

text/pvm_invocations.tex
```diff
@@ -80,30 +80,31 @@ \subsection{Refine Invocation}\label{sec:refineinvocation}
...
+      \Omega_\Gamma(\gascounter, \registers, \memory, \workingset, \context, w_\wi¬serviceindex, \accounts, (p_\wp¬context)_\wc¬lookupanchortime) &\when n = \mathtt{compile}\\
+      \Omega_M(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{machine}\\
+      \Omega_P(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{peek}\\
+      \Omega_O(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{poke}\\
+      \Omega_Z(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{pages}\\
+      \Omega_K(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{invoke}\\
+      \Omega_X(\gascounter, \registers, \memory, \workingset, \context) &\when n = \mathtt{expunge}\\
+      \tup{\oog, 0, \registers', \memory, \workingset, \context} &\otherwhen \gascounter < \Cgasunknown\\
+      \tup{\continue, \gascounter - \Cgasunknown, \registers', \memory, \workingset, \context} &\otherwise\\
+      \multicolumn{2}{l}{\where \context = \tup{\mathbf{m}, \pvm¬module¬map, \mathbf{e}}} \\
```
zdave-parity: Think this line is not necessary -- the individual context members are not used here?

text/pvm_invocations.tex
```diff
@@ -275,16 +277,16 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
...
 
-All host functions are of the form $\tup{\execst^* \in \set{\continue, \panic, \oog}, \gascounter^* \in \gas, \registers^* \in \regs, \memory^* \in \ram, \context^*} = \Omega_\square(\gascounter \in \gas, \registers \in \regs, \memory \in \ram, \context, \dots)$. The gas cost of a host-call is given by $g$, defined in the tables below; should there be insufficient gas to cover this cost, $\execst^* = \oog$ is returned, all remaining gas is consumed, and registers, memory and context are left unchanged. Formally:
+All host functions are of the form $\tup{\execst^* \in \set{\continue, \panic, \oog}, \gascounter^* \in \gas, \registers^* \in \regs, \memory^* \in \ram, \workingset^*, \context^*} = \Omega_\square(\gascounter \in \gas, \registers \in \regs, \memory \in \ram, \workingset \in \sequence{\optional{\pageid}}, \context, \dots)$. The gas cost of a host-call is given by $g$, defined in the tables below; should there be insufficient gas to cover this cost, $\execst^* = \oog$ is returned, all remaining gas is consumed, and registers, memory, working-set and context are left unchanged. Formally:
```
zdave-parity: Should add a type on `\workingset^*` as well I guess, like `\workingset^* \in \sequence{\optional{\pageid}}`.

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
 \end{equation}
 With base cost $c$, total gas is $c + \fnmemgas(L, \ell)$. The host-call tables below use this formula for all memory-sized terms.
 
-When transitioning \textsc{pvm} state for a host-call, a number of conditions typically hold true and host functions are defined essentially by their exceptions to these rules. Specifically, the machine does not halt, and \textsc{ram}, registers and context are unchanged. Formally:
```
zdave-parity: I would move this to just before the table to keep all the stuff that talks about gas together.

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
...
+    \wscharge¬extwrite(\registers, \memory, o, \mathbf{v}, f, l) &\equiv \begin{cases}
+      \tup{\panic, \registers_7, \memory} &\when \mathbf{v} = \error \vee \Nrange{o}{l} \not\subseteq \writable{\memory} \\
+      \tup{\continue, \mathtt{NONE}, \memory} &\otherwhen \mathbf{v} = \none \\
+      \tup{\continue, \len{\mathbf{v}}, \memory^\uparrow \exc (\memory^\uparrow)\subrange{o}{l} = \mathbf{v}\subrange{f}{l}} &\otherwise
+    \end{cases} \\
+    \where \memory^\uparrow &= \memory \exc (\memory^\uparrow_\ram¬frames)\subrange{\floor{\nicefrac{o}{\Cpvmpagesize}}}{\len{\spannedpages(o, l)}} = \sq{\frame¬present, \frame¬present, \dots}
+  \end{aligned}
+\end{equation}
+
+In the descriptions that follow, the working-set argument and result are elided except where a function threads them explicitly.
```
zdave-parity: Not sure what you mean by the "result" being elided? You mean like `\registers'_7` or something?

I would prefer to not elide the working-set argument, I think it's a bit weird to specifically exclude that?

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
...
+  \execst' \equiv \continue,\quad \registers' \equiv \registers,\quad \memory' \equiv \memory,\quad \workingset' \equiv \workingset,\quad \context' \equiv \context\qquad\text{except as indicated below}
 \end{equation}
 
+Host-call inspections and mutations of a machine's \textsc{ram} may carry a memory-hierarchy cost beyond the base and linear terms of the tables below, depending on what lies at the other end of the transfer. Such costs are terms of the host-call's $g$ and so are subject to the gas rule above. Where that end is itself a \textsc{pvm} machine the access is charged through the working-set as an ordinary \textsc{pvm} memory access would be. We define $\wscharge¬internal$, which computes this charge for a sequence $\mathbf{a}$ of octet accesses, each a pair of machine ($\none$ for the invoking machine) and address, and returns it together with the posterior working-set, \textsc{ram} and inner instances, in which the frame of each page touched is present:
+\begin{equation}
+  \begin{aligned}
+    \wscharge¬internal(w, \mem, \mathbf{m}, \mathbf{a}) &\equiv \begin{cases}
+      \tup{0, w, \mem, \mathbf{m}} &\when \mathbf{a} = \sq{} \\
```
zdave-parity: As before, the working-set is a composite type so to match convention we should use `\mathbf` for it.

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
...
+      \tup{g^\delta + g', w'', \mem'', \mathbf{m}''} &\otherwise
+    \end{cases} \\
+    \where \tup{m, x} &= \mathbf{a}_0\,,\quad i = \floor{\nicefrac{x}{\Cpvmpagesize}}\,,\quad \mu = \begin{cases} \mem &\when m = \none \\ \mathbf{m}\subb{m}_\pg¬ram &\otherwise \end{cases} \\
```
zdave-parity: `\mem` is `\mu` so we need a different name for the intermediate here...

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
...
+
+Octets \emph{read out of} a machine's \textsc{ram} neither provision a frame nor alter the working-set, and carry no charge beyond the host-call's $\fnmemgas$ term.
+
+Octets \emph{written into} a machine's \textsc{ram} likewise stay out of the working-set (so a page not already resident faults ordinarily on the machine's next access), however they \emph{do} provision a new frame where none yet exists, paying the allocation-and-fill cost a first-touch fault would, less the base $\Cpvmgaspagefault$ (which is deferred to that eventual fault). Such octets, having bypassed the working-set, are treated as non-resident when the machine reads them back, whether or not their page is resident; this read-back premium is not levied separately but folded into the per-octet cost of each delivering host-call, whose memory rate is accordingly raised by $\Cpvmgasdeliver$ (gas per 1024 octets). The provisioning cost of writing $l$ octets of a value $\mathbf{v}$ into the invoking machine's \textsc{ram} at $o$ is $\wscharge¬extwritegas$, a term of the host-call's $g$. Where the host-call panics or has nothing to write, no octets are written and the cost is zero. Formally:
```
zdave-parity: "Such octets, having bypassed the working-set, are treated as non-resident when the machine reads them back" is rather misleading I think. I would say that they are _treated_ as resident if they were in the working-set (as we don't change the working-set), but in practice we expect them to not be resident, or at least require pulling back from the other core's L3. We need to account for this disparity by adding `\Cpvmgasdeliver`.

text/pvm_invocations.tex
```diff
@@ -294,11 +296,57 @@ \subsection{Host Functions}
...
+\begin{equation}
+  \begin{aligned}
+    \wscharge¬extwritegas(\memory, o, \mathbf{v}, l) &\equiv \begin{cases}
+      0 &\when \mathbf{v} \in \set{\error, \none} \vee \Nrange{o}{l} \not\subseteq \writable{\memory} \\
```
zdave-parity: This is kind of weird, as it assumes you are using the `extwrite` logic below. I think we can just combine the two functions then there are no assumptions required?

text/pvm_invocations.tex
```diff
@@ -593,17 +649,18 @@ \subsubsection{Refine Functions}\label{sec:refinefunctions}
...
     \using \sq{n, o} &= \registers_{7, 8} \\
+    \using \arg &= \sq{\build{\tup{\none, x}}{x \orderedin \Nrange{o}{112}}} \\
     \using \tup{g_R, \mathbf{w}} &= \begin{cases}
       \tup{g_R, \mathbf{w}}: \encode[8]{g_R} \concat \encode[8]{\mathbf{w}} = \mem\subrange{o}{112} &\when \Nrange{o}{112} \subseteq \writable{\mem} \\
-      %\tup{\decode[8]{\memr\subrange{o}{8}}, \sq{\build{\decode[4]{\memr\subrange{o+8+8x}{8}}}{x \orderedin \Nmax{13}}}} &\when \Nrange{o}{60} \subset \writable_\mem} \\
       \tup{0, \error} &\otherwise
     \end{cases} \\
-    \using \tup{c, i', g_R', \pg¬gaschargedflag', \mathbf{w}', \mathbf{u}'} &= \Psi(\mathbf{m}\subb{n}_\pg¬code, \mathbf{m}\subb{n}_\pg¬pc, g_R, \mathbf{m}\subb{n}_{\pg¬gaschargedflag}, \mathbf{w}, \mathbf{m}\subb{n}_\pg¬ram)\\
-    \using \mem^* &= \mem \exc \mem^*\subrange{o}{112} = \encode[8]{g_R'} \concat \encode[8]{\mathbf{w}'}\\
+    \using \tup{g_1, \workingset_1, \mem_1, \mathbf{m}_1} &= \wscharge¬internal(\workingset, \mem, \mathbf{m}, \arg) \\
```
zdave-parity: We should avoid names like `\workingset_1` as a subscript on a sequence means index the sequence.

PR #533, opened 2026-06-02
Various host function fixes/improvements

zdave-parity:
The state after each commit should be reasonable, so this is probably best reviewed one commit at a time. Ignore merge commits and @arkpar's commits (which were reviewed in #517).

The only intentional behavioural change here is that, to be consistent with every other host function, `grow_heap` now returns `gascounter' = 0` in all out-of-gas cases. Previously there was one case where it returned out-of-gas and didn't change the gas counter. I assume this was an oversight.

PR #535, opened 2026-06-15
Fix typo in Accumulate Invocation result conversion

libingjiang47:
Fix branch result in the Accumulate Invocation collapse function.

https://graypaper.fluffylabs.dev/#/07f041d/383003383003?v=0.8.0

This appears to be a typo:

1. $(\mathbf{x}, \mathbf{y})_{\mathbf{p}}$ does not match the tuple-subscript convention.
2. According to the accumulation invocation semantics, this branch should output $\mathbf{x}_{\mathbf{p}}$.


gavofyork (approved):

zdave-parity:
Agree, I also spotted this and fixed it in #533 (which has just been merged).

Issue #529, opened 2026-05-27
Authority Services


gavofyork:
This is to allow services to set a privileged Authority service which can e.g. help rescuing the service should it be borked, provide for a trustless form of inter-service lending.

### Summary

All services $\mathbf{a}$ have two additional fields, the _Authority ID_ ($\mathbf{a}_u$) and the _Authority Balance_ ($\mathbf{a}_x$):

$\forall \mathbf{a} \in $\mathbb{A}: \mathbf{a}_u \in \mathbb{N}_S\mathsf{?}, \mathbf{a}_x \in \mathbb{N}_B$

The Authority ID field encodes the Authority of some *Junior* service, in case it is defined. When defined, all non-privileged mutable operations are permitted to be done on the Junior service by the Authority service.

In addition to the regular balance of a (Junior) service, there is an additional account field named *Authority Balance*, by default zero and aggregated with the regular balance when determining data footprint limits for the service. *This can only be transferred by the service holding sudo privileges of the account in question; or the service itself only when that is undefined.*

Mutable service operation host-calls (`write`, `assign`, `solicit`, `forget`, `new`, `upgrade`, `transfer` and `eject`) all take an optional extra `target` parameter, allowing the service on which the operation is conducted be specified. The host-call fails if `target` is specified and that target service's Authority is not the Executing Service.

The `transfer` host-call takes two additional binary flags to denote whether the amount to be transferred should be deducted from, and deposited into, the Authority Balance. The according Transfer Record also encodes the latter of these. Only the Authority is able to issue `transfer`s which deduct funds from the Authority Balance.

An additional host-call `authority(target?, value?)` allows a service with no defined authority to set its own Authority to a desired value (or retain it as undefined if `value` is null) when `target` is null. If `target` is specified, then the operation happens on the Target service, and it must already have an Authority set to the Executing Service.

In essence, a service is be able to set its Authority service from None to anything, but only the an Authority service is able to reset once a service's Authority has been set.

For simplicity, collisions in state resolution can happen by favouring Authority changes to a service.

### Multiplexer

We’ll probably want to prototype a multiplexer service too, which can be set as a target and then told to add Sudo services. Said services would send it messages which would just be sender-checked and then enacted via host calls on the target service.

gavofyork:
(Note added `assign` to the list of intermediatable host-calls, to allow an authority to transfer back any core-assignment token.)

PR #539, opened 2026-07-21
Junior services

zdave-parity:
All services have three new fields:

- Supervisor, which may be set to the service itself.
- Supervisor creation slot, see the explanation of this in accounts.tex.
- Supervisor balance.

A service's supervisor can mutate the service (write, solicit, forget,
upgrade host-calls) and can transfer from its balances.

A service cannot transfer from its supervisor balance (unless it is set
as the supervisor), however the supervisor balance is added to the
regular balance when determining storage limits for the service.

There is a new supervisor host-call to modify a service's supervisor.
This can only be called by the current supervisor.

The eject host-call has been simplified:

- It may now only be called by the ejected service's supervisor.
- It now requires that the ejected service's storage be _empty_.
  Previously eject could handle removal of a single preimage. This is
  not necessary any more as the supervisor can simply use the forget
  host-call to do this.

Note that it is now possible for a service to execute in the same
execution round in which it is ejected; the results of this execution
are simply discarded.

The transfer host-call now takes a source service argument and two flags
indicating whether to debit the source's regular or supervisor balance
and whether to credit the destination's regular or supervisor balance.

For convenience, the transfer host-call now supports immediate transfers
in some cases. An immediate transfer is requested by passing a null memo
pointer. No deferred transfer is created in this case.

The new host-call supports transferring the initial deposit from any
owned balance. The supervisor of the created service is initially set to
the calling service. This can be immediately changed using the
supervisor host-call if desired.

Note that to keep conflict resolution simple, a supervisor currently
_cannot_ use any global privileges assigned to its junior services.

See #529.


zdave-parity:
This is based on #533, all of the authority service stuff is in the last commit.

text/pvm_invocations.tex
```diff
@@ -93,20 +90,19 @@ \subsection{Refine Invocation}\label{sec:refineinvocation}
...
+      \Omega_H(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}, w_\wi¬serviceindex, \accounts, (p_\wp¬context)_\wc¬lookupanchortime) &\when n = \mathtt{historical\_lookup}\\
+      \Omega_E(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}, \segoff) &\when n = \mathtt{export}\\
+      \Omega_M(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{machine}\\
+      \Omega_P(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{peek}\\
+      \Omega_O(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{poke}\\
+      \Omega_Z(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{pages}\\
+      \Omega_K(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{invoke}\\
+      \Omega_X(\gascounter, \registers, \memory, \tup{\mathbf{m}, \mathbf{e}}) &\when n = \mathtt{expunge}\\
+      \tup{\oog, 0, \registers', \memory, \tup{\mathbf{m}, \mathbf{e}}} &\otherwhen \gascounter < \Cgasunknown\\
+      \tup{\continue, \gascounter - \Cgasunknown', \registers', \memory, \tup{\mathbf{m}, \mathbf{e}}} &\otherwise\\
```
davxy:
```suggestion
      \tup{\continue, \gascounter - \Cgasunknown, \registers', \memory, \tup{\mathbf{m}, \mathbf{e}}} &\otherwise\\
```

text/accounts.tex
```diff
@@ -179,3 +181,8 @@ \subsection{Service Privileges}
   \assigners &\in \sequence[\Ccorecount]{\serviceid} \ ,\qquad
   \alwaysaccers \in \dictionary{\serviceid}{\gas}
 \end{align}
+
+\subsection{Service Authorities}
+In addition to the globally privileged services discussed above, each service has a designated \emph{authority} service $\sa¬authority$; all non-privileged mutating operations are permitted to be done on a service by its authority service. Only the authority service may designate a new authority service. A service's authority service may be itself.
+
+In addition to the regular balance of a service $\sa¬balance$, there is an additional authority balance $\sa¬authbalance$, by default zero and aggregated with the regular balance when determining data footprint limits for the service. This can only be transferred by the authority service.
```
davxy: Nit. 2x "In addition", reformulate?

gavofyork:
There's a few conflicts...

text/accumulation.tex
```diff
@@ -243,29 +245,57 @@ \subsection{Execution}
...
+\begin{gather}
+  R_D(\mathbf{a}, \mathbf{w}, \mathbf{b}) \equiv \set{\build{\kv{\mathbf{k}}{\mathbf{v}}}{\kv{\mathbf{k}}{\mathbf{v}} \in \mathbf{a}, \mathbf{k} \in \mathbf{w}}} \cup (\mathbf{b} \setminus \mathbf{w}) \\
+  R_P(\mathbf{p}, \mathbf{l}) \equiv \set{\build{\kv{h}{\mathbf{p}\subb{h}}}{\kv{\tup{h, z}}{\mathbf{r}} \in \mathbf{l}, \mathbf{r} \ne \sq{}}} \\
+  R_A^\dagger(\mathbf{o}, \mathbf{a}, \mathbf{w}, \mathbf{r}, \mathbf{b}) \equiv \tup{\begin{aligned}
+    &\is{\sa¬codehash}{R(\mathbf{o}_\sa¬codehash, \mathbf{a}_\sa¬codehash, \mathbf{b}_\sa¬codehash)},
+      \is{\sa¬storage}{R_D(\mathbf{a}_\sa¬storage, \mathbf{w}, \mathbf{b}_\sa¬storage)}, \\
+    &\is{\sa¬requests}{R_D(\mathbf{a}_\sa¬requests, \mathbf{r}, \mathbf{b}_\sa¬requests)},
+      \is{\sa¬balance}{\mathbf{b}_\sa¬balance},
+      \is{\sa¬authority}{\mathbf{a}_\sa¬authority},
+      \is{\sa¬authbalance}{\mathbf{a}_\sa¬authbalance}, \\
```
davxy: Taking $\mathbf{a}_x$ alone drops deferred transfers with destauth=⊤ : $Ψ_A$ credits them into the destination's accumulation context, so they only appear in $\mathbf{b}_x$ and thus the authority's context never sees them (it only receives transfers destined to itself). So when a service's authority is not itself, the credit is discarded at the merge while the sender's debit (made at transfer call time) persists.

Now we have $x = \mathbf{a}_x$ which is  only the authority's view; deferred credits (which live in $\mathbf{b}_x$) are lost.
A fix can be $x = a_x + b_x - o_x$ : the authority's view plus the delta from the service's own run, which can only be deferred credits. If nothing the $\mathbf{b}_x = \mathbf{o}_x$

zdave-parity:
Rebased, no changes.

text/accumulation.tex
zdave-parity: Hmm yeah thanks, good spot. Note that `a` and `b` can be from the same service so `a_x + b_x - o_x` doesn't work... in the case they are the same you would double any change! I can of course detect that case and make it work. Perhaps transfers to the authority balance should be processed by the authority service anyway though? Any opinion here @gavofyork?

text/accounts.tex
zdave-parity: Yeah whoops. The first "in addition" is fluff anyway, I've just removed it.

text/accumulation.tex
davxy: Just for clarity... Funds are lost exactly when (in `transfer` hostcall) $\mathbf{o} \neq 0$ and $x \in {2,3}$ and dest's authority not equal dest: the `transfer` hostcall enqueues a deferred transfer for `dest`, which is delivered to dest's own accumulation (so credit lands in **b**), but R_A† takes authbalance from **a**, which never sees the transfer

zdave-parity: Should be fixed now. I've stuck with transfers to the authority balance being handled by the service rather than its authority service. The authority balance deltas are simply added as you proposed, with a check in $R_A$ to handle the case where the service and its authority service are actually the same.

gavofyork: > Hmm yeah thanks, good spot. Note that `a` and `b` can be from the same service so `a_x + b_x - o_x` doesn't work... in the case they are the same you would double any change! I can of course detect that case and make it work. Perhaps transfers to the authority balance should be processed by the authority service anyway though? Any opinion here @gavofyork?

As discussed I think it makes more sense to keep all transfers be processed by the destination service.

text/definitions.tex
```diff
@@ -331,6 +332,7 @@ \subsubsection{Host-function gas costs}
   \item[$\CgasS = 2193$] $\Omega_S$ (\texttt{solicit}) base gas cost.
   \item[$\CgasT = 575$] $\Omega_T$ (\texttt{transfer}) base gas cost.
   \item[$\CgasU = 1028$] $\Omega_U$ (\texttt{upgrade}) base gas cost.
+  \item[$\CgasV = 1000$] $\Omega_V$ (\texttt{authority}) base gas cost.
```
davxy: I imagine that 1000 is a placeholder?

text/pvm_invocations.tex
```diff
@@ -787,54 +816,99 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
-      \multicolumn{2}{l}{\quad \where \mathbf{d} = \set{ \kv{\local¬desiredid}{\mathbf{a}}, \kv{\imX_\im¬id}{\mathbf{s}} }}\\
-      \tup{\continue, \imX_\im¬nextfreeid, i^*, (\imX_\im¬state)_\ps¬accounts \cup \mathbf{d}} &\otherwise \\
-      \multicolumn{2}{l}{\quad \where i^* = \text{check}(\Cminpublicindex + (\imX_\im¬nextfreeid - \Cminpublicindex + 42) \bmod (2^{32} - \Cminpublicindex - 2^8))}\\
-      \multicolumn{2}{l}{\quad \also \mathbf{d} = \set{ \kv{\imX_\im¬nextfreeid}{\mathbf{a}}, \kv{\imX_\im¬id}{\mathbf{s}} }}\\
+      \tup{\panic, \registers_7, \imX_\im¬nextfreeid, \mathbf{d}} &\when \Nrange{o}{60} \not\subseteq \readable{\memory} \\
+      \tup{\continue, \mathtt{WHO}, \imX_\im¬nextfreeid, \mathbf{d}} &\otherwhen s \not\in \mathbf{d} \\
+      \tup{\continue, \mathtt{HUH}, \imX_\im¬nextfreeid, \mathbf{d}} &\otherwhen x > 3 \lor \imX_\im¬id \ne p \lor a < \mathbf{a}_\sa¬minbalance \lor (\sa¬gratis \ne 0 \wedge \imX_\im¬id \ne (\imX_\im¬state)_\ps¬manager) \\
+      \tup{\continue, \mathtt{CASH}, \imX_\im¬nextfreeid, \mathbf{d}} &\otherwhen \mathbf{d}'\subb{s}_\sa¬balance < 0 \lor \mathbf{d}'\subb{s}_\sa¬authbalance < 0 \lor {} \\
```
davxy: The `CASH` condition tests $d'[s]_b < 0 ∨ d'[s]_x < 0$, but both fields are defined in the account struct as  unsigned integers. $d'[s]_b$ above is defined as $d[s]_b − a$, which (strictly speaking) can lead to UB.

Potential solution: `CASH` when $(x ∈ {0,2} ∧ a > d[s]_b) ∨ (x ∈ {1,3} ∧ a > d[s]_x)$ or maybe track the debit as an integers

Same issue exists in the `transfer` hostcall

Of course, everyone will likely implement this correctly (i.e. first check for underflows), but it may be worth formalize it to avoid UB.

text/pvm_invocations.tex
```diff
@@ -151,37 +153,44 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
...
         \is{\ao¬poststate}{\local¬postxferstate},
+        \is{\ao¬writtenkeys}{\emset},
+        \is{\ao¬modifiedreqs}{\emset},
         \is{\ao¬defxfers}{\sq{}},
         \is{\ao¬yield}{\none},
         \is{\ao¬gasused}{0},
         \is{\ao¬provisions}{\sq{}}
       }
-        &\when \jam¬blob = \none \vee \len{\jam¬blob} > \Cmaxservicecodesize \\
+        &\when s \not\in \local¬basestate_\ps¬accounts \vee \jam¬blob = \none \vee \len{\jam¬blob} > \Cmaxservicecodesize \\
```
davxy: Perhaps $s \not\in \textbf{e}_\textbf{d}$ is not a sufficient guard :-/
In this case, the **e** in the returned tuple is set to **s**, with **s** defined below as $\textbf{s}[s] = ...$ , i.e. we perform indirection using s itself, which is precisely the condition we're trying to guard against.

Maybe we can split in two cases?
- if $s \not\in \textbf{e}_\textbf{d}$ then we return a tuple that keeps **e** untouched
- if $𝔧 = ∅ or |𝔧| > W_C$ we do the **e**= **s** (this is necessary as the service exists so transfer credits needs to land)

text/pvm_invocations.tex
```diff
@@ -843,55 +917,71 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
-      \tup{\continue, \mathtt{WHO}, (\imX_\im¬state)_\ps¬accounts} &\otherwhen \mathbf{d} = \error \vee \mathbf{d}_\sa¬codehash \ne \encode[32]{\imX_\im¬id} \\
-      \tup{\continue, \mathtt{HUH}, (\imX_\im¬state)_\ps¬accounts} &\otherwhen \mathbf{d}_\sa¬items \ne 2 \vee \tup{h, l} \not\in \mathbf{d}_\sa¬requests \\
-      \tup{\continue, \mathtt{OK}, (\imX_\im¬state)_\ps¬accounts \setminus \set{d} \cup \set{ \kv{\imX_\im¬id}{\mathbf{s}'} }} &\otherwhen \mathbf{d}_\sa¬requests\subb{h, l} = \sq{x, y}, y < t - \Cexpungeperiod \\
-      \tup{\continue, \mathtt{HUH}, (\imX_\im¬state)_\ps¬accounts} &\otherwise \\
+    \using d &= \registers_7 \\
+    \using \mathbf{d} &= (\imX_\im¬state)_\ps¬accounts\subb{d} \\
+    \using \mathbf{s}' &= \imX_\im¬self \exc \mathbf{s}'_\sa¬balance = (\imX_\im¬self)_\sa¬balance + \mathbf{d}_\sa¬balance + \mathbf{d}_\sa¬authbalance \\
+    \tup{\registers'_7, (\imX'_\im¬state)_\ps¬accounts} &\equiv \begin{cases}
+      \tup{\mathtt{WHO}, (\imX_\im¬state)_\ps¬accounts} &\when \mathbf{d} = \none \\
+      \tup{\mathtt{HUH}, (\imX_\im¬state)_\ps¬accounts} &\otherwhen \imX_\im¬id \ne \mathbf{d}_\sa¬authority \vee \mathbf{d}_\sa¬items \ne 0 \\
```
davxy: Looks like `𝐝_i ≠ 0` alone is not sufficient now:

At slot N, a service calls `forget(self, c, z)` on its own code, (code preimage requests change: `[x] → [x, N]`), but the service is still functional since its code is `𝐚_𝐩[𝐚_c]`.

Then at slot `t > N + D`, in a single accumulation:
1. `write(self, k, ∅)` for all keys => `i = 2`, with `i = 2|𝐥| + |𝐬|`
2. `forget(self, c, z)`, now request and preimage dropped, `i = 0`
3. `eject` on its own index, legal since a service may be its own authority. Result is `(𝐱_𝐞)_𝐝 ∖ {d} ∪ {𝐱_s ↦ 𝐬'}`, so for `d = 𝐱_s` the union undoes the removal and the service survives with `𝐬'_b = 𝐬_b + 𝐝_b + 𝐝_x`, i.e. `b' = 2b + x`
4. `solicit(self, c, z)` to re-request its code. Preimage re-suppliable via extrinsic later

One more `forget` re-arms it, so that's a doubling every `D`.

The fix can be in `Ψ_A` and check `Λ(𝐚, t, 𝐚_c) ≠ ∅` rather than `𝐚_𝐜 ≠ ∅`, which kills it at the first `forget`. I.e. the service can't be called anymore in the accumulate calls, which seems the right thing to do.

davxy:
Design question. Nothing clears the authority field of services pointing at the one being ejected.
So if S has U as its authority and U gets ejected, S is just left with a dangling authority?

Probably fine in most cases, but ids get recycled. For the low range the registrar can create a new service at exactly U's old index, and that one silently becomes S's authority. Perhaps this can be exploited maliciously

text/definitions.tex
zdave-parity: Indeed.

text/pvm_invocations.tex
zdave-parity: My reasoning was that we don't claim that `d'` is a member of any particular set, so there is nothing contradictory here. In general intermediate values aren't explicitly constrained in the GP; implementations need to take care that whatever concrete types are used can fit the possible values. I think it would be less clear to write the condition as proposed?

text/pvm_invocations.tex
zdave-parity: Yeah agree. I didn't realise that `\mathbf{s}` was used in the first case, though it does make sense that we want to add transfers to our balance even if we don't have code to execute. Probably this should be reformatted to make it clear that `\mathbf{s}` is used in both cases...

zdave-parity:
> Design question. Nothing clears the authority field of services pointing at the one being ejected. So if S has U as its authority and U gets ejected, S is just left with a dangling authority?

Yes. Not sure we can realistically protect against this as we only track the relationship in the junior state; when ejecting a service we in general have no idea what junior services it has. We could of course track this, but it doesn't seem worthwhile just to protect against this happening.

> Probably fine in most cases, but ids get recycled. For the low range the registrar can create a new service at exactly U's old index, and that one silently becomes S's authority. Perhaps this can be exploited maliciously

I think the answer is "don't do this". You can also eject privileged services and their privileges can potentially then be picked up by newly created services. Again I think the answer is "don't do this". Of course at least with privileged services they can be reassigned by the manager service if something like that happens.

davxy:
> > Design question. Nothing clears the authority field of services pointing at the one being ejected. So if S has U as its authority and U gets ejected, S is just left with a dangling authority?
> 
> Yes. Not sure we can realistically protect against this as we only track the relationship in the junior state; when ejecting a service we in general have no idea what junior services it has. We could of course track this, but it doesn't seem worthwhile just to protect against this happening.

In the service account we have the creation timeslot. As a reference to authorities, perhaps we can use not just the authority service id, but the tuple `(id, timeslot)`, where `timeslot` is the authority's service creation timeslot. This would allow the hostcalls to detect whether the authority has changed, since a reincarnation would have the same id but a different creation timeslot. This is a simple change, but not sure if is worth BTW, maybe we can keep as-is for now?

text/pvm_invocations.tex
zdave-parity: Hmm yeah I thought the code would not be found after the first `forget`. But apparently I did not check this and it is not the case... it is available right up until the point it is actually evicted from the state. The `lookup` host-call also permits lookup of preimages which are not "available".

In any case I am going to change the conflict resolution behaviour so _all_ changes performed by the junior service are discarded if there are any modifications by the authority service. This can include ejection so there won't be any need to prevent the junior service accumulating in the same slot in which it is ejected.

zdave-parity:
> In the service account we have the creation timeslot. As a reference to authorities, perhaps we can use not just the authority service id, but the tuple `(id, timeslot)`, where `timeslot` is the authority's service creation timeslot. This would allow the hostcalls to detect whether the authority has changed, since a reincarnation would have the same id but a different creation timeslot. This is a simple change, but not sure if is worth BTW, maybe we can keep as-is for now?

Perhaps that plus allowing a service to change its authority in the old-authority-ejected case would be useful? As at the moment you cannot fix a dangling authority at all.

davxy:
> allowing a service to change its authority in the old-authority-ejected case would be useful

This would be definitely useful

text/pvm_invocations.tex
zdave-parity: Hopefully fixed now.

text/pvm_invocations.tex
zdave-parity: This should not be an issue any more.

zdave-parity:
> Perhaps that plus allowing a service to change its authority in the old-authority-ejected case would be useful? As at the moment you cannot fix a dangling authority at all.

Ok I've done this. Actually when a service's authority service is ejected the authority service implicitly reverts to the service itself.

Also changed conflict resolution as discussed on Matrix: we allow report accumulation to be done concurrently with authority service accumulation and simply discard the service's accumulation results in the case of a conflict. Transfer accumulation on the other hand is _never_ done concurrently with authority service accumulation.

gavofyork (changes requested):
Just one minor one so far - let's rename *authority* to something less overused in this field. "Senior", "controller" or "supervisor", perhaps.

gavofyork (dismissed):
Can't see anything obvious beyond the function refactor and the Account fields description.

zdave-parity:
> Just one minor one so far - let's rename _authority_ to something less overused in this field. "Senior", "controller" or "supervisor", perhaps.

I went with "supervisor".

> Can't see anything obvious beyond the function refactor and the Account fields description.

Changed the function to a slightly magic derived field as discussed. Split the odd tuple field into two regular fields. Hopefully improved the explanation of the logic behind tracking the creation slot of the supervisor.

I think everything has been addressed now...

text/accumulation.tex
```diff
@@ -148,6 +152,26 @@ \subsection{Execution}
...
+As the result of accumulating a transfer cannot reasonably be discarded (this would cause funds to be lost), deferred transfers to a service are \emph{never} accumulated in parallel with execution of the Accumulate logic of the supervisor. We define the function $T$ which partitions a sequence of deferred transfers into those which may be accumulated in parallel with the Accumulate logic of the services $\mathbf{s}$ and those which must be accumulated in a later round.
+\begin{equation}
+  \local¬fnpartitionxfers\colon\abracegroup{
+    &\tuple{\sequence{\defxfer}, \protoset{\serviceid}, \dictionary{\serviceid}{\serviceaccount}} \to \tuple{\sequence{\defxfer}, \sequence{\defxfer}} \\
+    &\tup{\sq{}, \mathbf{s}, \mathbf{d}} \mapsto \tup{\sq{}, \sq{}} \\
+    &\tup{\sq{t} \concat \mathbf{t}, \mathbf{s}, \mathbf{d}} \mapsto \begin{cases}
+      \tup{\mathbf{t}^!, \sq{t} \concat \mathbf{t}^Q} &\when d = \top \\
+      \tup{\sq{t} \concat \mathbf{t}^!, \mathbf{t}^Q} &\otherwise
+    \end{cases} \\
+    &\quad\where d = t_\dx¬dest \in \mathbf{d} \land \effectivesupervisor{\mathbf{d}}{t_\dx¬dest} \in \mathbf{s} \setminus \set{t_\dx¬dest} \\
```
zdave-parity: This check is not sufficient... `t_d` might be the supervisor of some service in `\mathbf{s}`, this only checks if `t_d`'s supervisor is in `\mathbf{s}`.

zdave-parity: Fixed...

text/accumulation.tex
```diff
@@ -187,12 +216,28 @@ \subsection{Execution}
...
+      N(\mathbf{o}) \in \acconeout &\equiv \tup{
+        \ao¬poststate,
+        \is{\ao¬modifiedaccounts}{\emset},
+        \is{\ao¬defxfers}{\sq{}},
+        \is{\ao¬yield}{\none},
+        \is{\ao¬gasused}{\mathbf{o}_\ao¬gasused},
+        \is{\ao¬provisions}{\mathbf{o}_\ao¬provisions}
+      } \\
+      \local¬acc(s) &\equiv \begin{cases}
+        N(\mathbf{o}) &\when s \in \mathbf{o}_\ao¬modifiedaccounts \land s \in \mathbf{y}_\ao¬modifiedaccounts \setminus \set{\effectivesupervisor{\ps¬accounts}{s}} \\
```
davxy, line 238:
```suggestion
        \mathbf{o} &\when s \not\in \keys{\ps¬accounts} \\
        N(\mathbf{o}) &\otherwhen s \in \mathbf{o}_\ao¬modifiedaccounts \land s \in \mathbf{y}_\ao¬modifiedaccounts \setminus \set{\effectivesupervisor{\ps¬accounts}{s}} \\
```

text/accumulation.tex
```diff
@@ -148,6 +152,30 @@ \subsection{Execution}
   \local¬fngasused \equiv \sequence{\tuple{\serviceid, \gas}}
 \end{equation}
 
+It is possible for a service's state to be modified by both the service's Accumulate logic and its supervisor's Accumulate logic. We wish to avoid inconsistent state resulting from concurrent modification while not unnecessarily restricting parallel accumulation. To this end, we optimistically permit execution of a service's Accumulate logic and its supervisor's Accumulate logic in the same round, but discard all results\footnote{With one exception: provided preimages are not discarded} of the service's execution if both modify the service's state. Note that ejection of a service is considered a modification of its state.
```
davxy, line 155:
```suggestion
It is possible for a service's state to be modified by both the service's Accumulate logic and its supervisor's Accumulate logic. We wish to avoid inconsistent state resulting from concurrent modification while not unnecessarily restricting parallel accumulation. To this end, we optimistically permit execution of a service's Accumulate logic and its supervisor's Accumulate logic in the same round, but discard all results\footnote{With one exception: provided preimages are not discarded} of the service's execution if both modify the service's state. Note that ejection of a service is considered a modification of its state. A service scheduled for accumulation may itself have been ejected in an earlier round, and thus no longer exist. Such a service accumulates to nothing and consumes no gas.
```

Nit for the new guard https://github.com/gavofyork/graypaper/pull/539#discussion_r3740341753

text/pvm_invocations.tex
```diff
@@ -140,48 +141,60 @@ \subsection{Accumulate Invocation}\label{sec:accumulateinvocation}
...
     \tup{\local¬basestate, t, s, g, \mathbf{i}} &\mapsto \begin{cases}
+      \tup{
+        \ao¬poststate,
+        \is{\ao¬modifiedaccounts}{\emset},
+        \is{\ao¬defxfers}{\sq{}},
+        \is{\ao¬yield}{\none},
+        \is{\ao¬gasused}{0},
+        \is{\ao¬provisions}{\sq{}}
+      }
+        &\when s \not\in \local¬basestate_\ps¬accounts \\
```
davxy: Deferred transfers are delivered a round after they're issued, so if the destination is ejected in between, the transfer is dropped with the source already debited and the amount destroyed. 

The issue is pre-existing in principle (and AFAIK known and accepted), but with Ψ_A's new first case is what makes it well-defined rather than an undefined lookup. The now supervisor-driven ejection makes it much easier to hit than the old.

If this is formally accepted, I think it would be beneficial to add a note about this potential funds-burning scenario in the accumulation section.

text/accumulation.tex
```diff
@@ -187,12 +216,28 @@ \subsection{Execution}
...
+        \is{\ao¬yield}{\none},
+        \is{\ao¬gasused}{\mathbf{o}_\ao¬gasused},
+        \is{\ao¬provisions}{\mathbf{o}_\ao¬provisions}
+      } \\
+      \local¬acc(s) &\equiv \begin{cases}
+        N(\mathbf{o}) &\when s \in \mathbf{o}_\ao¬modifiedaccounts \land s \in \mathbf{y}_\ao¬modifiedaccounts \setminus \set{\effectivesupervisor{\ps¬accounts}{s}} \\
+        \mathbf{o} &\otherwise
+      \end{cases} \\
+      &\phantom{{}\equiv{}} \where \mathbf{o} = \accone(\psX, \mathbf{t}, \mathbf{r}, \mathbf{f}, s) \\
+      &\phantom{{}\equiv{}} \also \mathbf{y} = \accone(\psX, \mathbf{t}, \mathbf{r}, \mathbf{f}, \effectivesupervisor{\ps¬accounts}{s}) \\
```
davxy, line 242: The "nullification" test uses the supervisor's raw execution, not its post-discard one. E.g. with e(a)=b, e(b)=c: a writes a, b writes a and b, c writes b , then b is nullified by c, but a is still nullified by b's write to a, which never lands.  I imagine that this is deliberate and is not worth over-complicating

text/accumulation.tex
```diff
@@ -3,6 +3,7 @@
 \newcommand*{\local¬fngasused}{U}
 \newcommand*{\local¬fnprovidable}{Y}
 \newcommand*{\local¬fnprovide}{I}
+\newcommand*{\local¬fnpartitionxfers}{T}
```
davxy, line 6: Nit. $T()$ collides with the statistics function in the same section.
Same for $N()$, but I think this is fine as it is defined within the context of $\Delta_*$

text/accounts.tex
```diff
@@ -179,3 +182,20 @@ \subsection{Service Privileges}
...
+
+Alongside a service's supervisor $\sa¬supervisor$, we record the time slot in which the supervisor was created $\sa¬supervisorcreated$. We define the \emph{effective supervisor} of a service to be the designated supervisor $\sa¬supervisor$ if this service exists and has the expected creation slot, or the service itself otherwise. The purpose of the creation slot check is to prevent unintentional transfer of power in the case that the supervisor is ejected and a new service is created with the same identifier: the new service will \emph{not} be considered the supervisor despite having the correct identifier, as it will have a newer creation slot. For convenience, given a service accounts dictionary $\mathbf{d}$ and a service identifier $s$, we express the effective supervisor of $s$ as $\effectivesupervisor{\mathbf{d}}{s}$. Formally:
+\begin{equation}\label{eq:effectivesupervisor}
+  \begin{aligned}
+    \forall \mathbf{d} \in \dictionary{\serviceid}{\serviceaccount}, s \in \keys{\mathbf{d}} : \effectivesupervisor{\mathbf{d}}{s} &\equiv \begin{cases}
+      u &\when u \in \mathbf{d} \land \mathbf{d}\subb{u}_\sa¬created = d \\
```
davxy:
```suggestion
      u &\when u \in \keys{\mathbf{d}} \land \mathbf{d}\subb{u}_\sa¬created = d \\
```

nit. Strictly speaking we need to check the set of keys of the dictionary. This is the convention we used so far. 

I drop a note in the other instance that I identified

text/accumulation.tex
```diff
@@ -148,6 +152,30 @@ \subsection{Execution}
...
+  \begin{gathered}
+    \local¬fnpartitionxfers(\mathbf{t}, \mathbf{s}, \mathbf{d}) \equiv \local¬fnpartitionxfers^*(\mathbf{t}, \mathbf{s}, \emptyset, \mathbf{d}) \\
+    \local¬fnpartitionxfers^*\colon\abracegroup{
+      &\tuple{\sequence{\defxfer}, \protoset{\serviceid}, \protoset{\serviceid}, \dictionary{\serviceid}{\serviceaccount}} \to \tuple{\sequence{\defxfer}, \sequence{\defxfer}} \\
+      &\tup{\sq{}, \mathbf{s}, \mathbf{i}, \mathbf{d}} \mapsto \tup{\sq{}, \sq{}} \\
+      &\tup{\sq{t} \concat \mathbf{t}, \mathbf{s}, \mathbf{i}, \mathbf{d}} \mapsto \begin{cases}
+        \tup{\mathbf{t}^!, \sq{t} \concat \mathbf{t}^Q} &\when q = \top \\
+        \tup{\sq{t} \concat \mathbf{t}^!, \mathbf{t}^Q} &\otherwise
+      \end{cases} \\
+      &\quad\where q = (t_\dx¬dest \in \mathbf{d} \land \effectivesupervisor{\mathbf{d}}{t_\dx¬dest} \in (\mathbf{s} \cup \mathbf{i}) \setminus \set{t_\dx¬dest}) \lor {} \\
```
davxy:
```suggestion
      &\quad\where q = (t_\dx¬dest \in \keys{\mathbf{d}} \land \effectivesupervisor{\mathbf{d}}{t_\dx¬dest} \in (\mathbf{s} \cup \mathbf{i}) \setminus \set{t_\dx¬dest}) \lor {} \\
```

text/accumulation.tex
```diff
@@ -148,6 +152,30 @@ \subsection{Execution}
...
+      &\quad\phantom{\where q = {}} \exists s \in \mathbf{i} \setminus \set{t_\dx¬dest} : \effectivesupervisor{\mathbf{d}}{s} = t_\dx¬dest \\
+      &\quad\also \mathbf{i}' = \begin{cases}
+        \mathbf{i} &\when q = \top \lor t_\dx¬dest \not\in \mathbf{d} \\
```
davxy:
```suggestion
        \mathbf{i} &\when q = \top \lor t_\dx¬dest \not\in \keys{\mathbf{d}} \\
```

text/pvm_invocations.tex
```diff
@@ -787,54 +817,107 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
-      \tup{\continue, \imX_\im¬nextfreeid, i^*, (\imX_\im¬state)_\ps¬accounts \cup \mathbf{d}} &\otherwise \\
-      \multicolumn{2}{l}{\quad \where i^* = \text{check}(\Cminpublicindex + (\imX_\im¬nextfreeid - \Cminpublicindex + 42) \bmod (2^{32} - \Cminpublicindex - 2^8))}\\
-      \multicolumn{2}{l}{\quad \also \mathbf{d} = \set{ \kv{\imX_\im¬nextfreeid}{\mathbf{a}}, \kv{\imX_\im¬id}{\mathbf{s}} }}\\
+    \using p &= \begin{cases}
+      \imX_\im¬id \in \set{s, \effectivesupervisor{\mathbf{d}}{s}} &\when x \in \set{0, 2} \\
+      \imX_\im¬id = \effectivesupervisor{\mathbf{d}}{s} &\otherwise
+    \end{cases} \\
+    \tup{\execst', \registers'_7, \imX'_\im¬nextfreeid, (\imX'_\im¬state)_\ps¬accounts, \imX'_\im¬modifiedaccounts} &\equiv \begin{cases}
+      \tup{\panic, \registers_7, \imX_\im¬nextfreeid, \mathbf{d}, \imX_\im¬modifiedaccounts} &\when \Nrange{o}{60} \not\subseteq \readable{\memory} \\
+      \tup{\continue, \mathtt{WHO}, \imX_\im¬nextfreeid, \mathbf{d}, \imX_\im¬modifiedaccounts} &\otherwhen s \not\in \mathbf{d} \\
```
davxy:
```suggestion
      \tup{\continue, \mathtt{WHO}, \imX_\im¬nextfreeid, \mathbf{d}, \imX_\im¬modifiedaccounts} &\otherwhen s \not\in \keys{\mathbf{d}} \\
```

text/pvm_invocations.tex
```diff
@@ -875,23 +984,17 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
-      \tup{\continue, \mathtt{OK}, (\imX_\im¬state)_\ps¬accounts \setminus \set{d} \cup \set{ \kv{\imX_\im¬id}{\mathbf{s}'} }} &\otherwhen \mathbf{d}_\sa¬requests\subb{h, l} = \sq{x, y}, y < t - \Cexpungeperiod \\
-      \tup{\continue, \mathtt{HUH}, (\imX_\im¬state)_\ps¬accounts} &\otherwise \\
+    \using d &= \registers_7 \\
+    \using \mathbf{d} &= (\imX_\im¬state)_\ps¬accounts \\
+    \using \mathbf{d}' &= \mathbf{d} \exc \abracegroup{
+      &d \not\in \keys{\mathbf{d}'} \\
+      &\mathbf{d}'\subb{\imX_\im¬id}_\sa¬balance = (\imX_\im¬self)_\sa¬balance + \mathbf{d}\subb{d}_\sa¬balance + \mathbf{d}\subb{d}_\sa¬supervisorbalance \\
+    } \\
+    \tup{\registers'_7, (\imX'_\im¬state)_\ps¬accounts, \imX'_\im¬modifiedaccounts} &\equiv \begin{cases}
+      \tup{\mathtt{WHO}, \mathbf{d}, \imX_\im¬modifiedaccounts} &\when d \not\in \mathbf{d} \\
```
davxy:
```suggestion
      \tup{\mathtt{WHO}, \mathbf{d}, \imX_\im¬modifiedaccounts} &\when d \not\in \keys{\mathbf{d}} \\
```

text/definitions.tex
davxy:
```suggestion
  \item[$\Cmaxservicecodesize = 4,000,000$] The maximum size of service code in octets. See equations \ref{eq:refinvocation} \& \ref{eq:accinvocation}.
```

I noticed `onxferinvocation` dangling reference. Perhaps we can just remove it here

davxy (approved):
Overall LGTM. Mostly nits and considerations.

IMO the only part I identified that warrants a fix is:  https://github.com/gavofyork/graypaper/pull/539#discussion_r3740341753

text/pvm_invocations.tex
```diff
@@ -787,54 +817,107 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
   $\begin{aligned}
-    \using \sq{o, l, \sa¬minaccgas, \sa¬minmemogas, \sa¬gratis, \local¬desiredid} &= \registers\subrange{7}{6} \\
-    \using \sa¬codehash &= \begin{cases}
-      \memory\subrange{o}{32} &\when \Nrange{o}{32} \subseteq \readable{\memory} \wedge l \in \Nbits{32} \\
-      \error &\otherwise
+    \using s &= \begin{cases}
+      \imX_\im¬id &\when \registers_7 = 2^{64} - 1 \\
+      \registers_7 &\otherwise
+    \end{cases} \\
+    \using \sq{\local¬desiredid, x, o} &= \registers\subrange{8}{3} \\
```
davxy, line 824: nit. x is bounded twice in this block. The flags register and the new account's supervisor balance. Perhaps we can use a different letter for the flags register?

text/accumulation.tex
```diff
@@ -148,6 +152,30 @@ \subsection{Execution}
   \local¬fngasused \equiv \sequence{\tuple{\serviceid, \gas}}
 \end{equation}
 
+It is possible for a service's state to be modified by both the service's Accumulate logic and its supervisor's Accumulate logic. We wish to avoid inconsistent state resulting from concurrent modification while not unnecessarily restricting parallel accumulation. To this end, we optimistically permit execution of a service's Accumulate logic and its supervisor's Accumulate logic in the same round, but discard all results\footnote{With one exception: provided preimages are not discarded} of the service's execution if both modify the service's state. Note that ejection of a service is considered a modification of its state.
+
+As the result of accumulating a transfer cannot reasonably be discarded (this would cause funds to be lost), deferred transfers to a service are \emph{never} accumulated in parallel with execution of the Accumulate logic of the supervisor. We define the function $T$ which partitions a sequence of deferred transfers into those which may be accumulated in parallel with the Accumulate logic of the services $\mathbf{s}$ and those which must be accumulated in a later round.
```
davxy:
```suggestion
As the result of accumulating a transfer cannot reasonably be discarded (this would cause funds to be lost), deferred transfers to a service are \emph{never} accumulated in parallel with execution of the Accumulate logic of the supervisor. We define the function $\local¬fnpartitionxfers$ which partitions a sequence of deferred transfers into those which may be accumulated in parallel with the Accumulate logic of the services $\mathbf{s}$ and those which must be accumulated in a later round.
```

text/pvm_invocations.tex
```diff
@@ -843,30 +926,56 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
+    \using \mathbf{d} &= (\imX_\im¬state)_\ps¬accounts \\
+    \using \mathbf{d}' &= \mathbf{d} \exc \begin{cases}
+      \mathbf{d}'\subb{\dx¬source}_\sa¬balance = \mathbf{d}\subb{\dx¬source}_\sa¬balance - \dx¬amount &\when x \in \set{0, 2} \\
+      \mathbf{d}'\subb{\dx¬source}_\sa¬supervisorbalance = \mathbf{d}\subb{\dx¬source}_\sa¬supervisorbalance - \dx¬amount &\when x \in \set{1, 3} \\
+      \mathbf{d}'\subb{\dx¬dest}_\sa¬balance = \mathbf{d}\subb{\dx¬dest}_\sa¬balance + \dx¬amount &\when o = 0 \land x \in \set{0, 1} \\
+      \mathbf{d}'\subb{\dx¬dest}_\sa¬supervisorbalance = \mathbf{d}\subb{\dx¬dest}_\sa¬supervisorbalance + \dx¬amount &\when o = 0 \land x \in \set{2, 3}
+    \end{cases} \\
+    \using \mathbf{w} &= \begin{cases}
+      \set{s, d} &\when o = 0 \\
+      \set{s} &\otherwise
```
davxy:
```suggestion
      \set{\dx¬source, \dx¬dest} &\when o = 0 \\
      \set{\dx¬source} &\otherwise
```

text/pvm_invocations.tex
```diff
@@ -843,30 +926,56 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
...
+    \end{cases} \\
+    \using p_s &= \begin{cases}
+      \imX_\im¬id \in \set{\dx¬source, \effectivesupervisor{\mathbf{d}}{\dx¬source}} &\when x \in \set{0, 2} \\
+      \imX_\im¬id = \effectivesupervisor{\mathbf{d}}{\dx¬source} &\otherwise
+    \end{cases} \\
+    \using p_d &= o \ne 0 \lor (\imX_\im¬id \in \set{\dx¬dest, \effectivesupervisor{\mathbf{d}}{\dx¬dest}} \land (x \in \set{1, 2} \lor s \ne d)) \\
```
davxy:
```suggestion
    \using p_d &= o \ne 0 \lor (\imX_\im¬id \in \set{\dx¬dest, \effectivesupervisor{\mathbf{d}}{\dx¬dest}} \land (x \in \set{1, 2} \lor \dx¬source \ne \dx¬dest)) \\
```

davxy (dismissed):

text/accumulation.tex
zdave-parity: I didn't bother with an explicit check because `s \in \mathbf{o}_\ao¬modifiedaccounts` is well-defined but always false when the service `s` does not exist.

text/pvm_invocations.tex
zdave-parity: Hmm yeah ok, I'll add a paragraph about this somewhere.

text/accumulation.tex
zdave-parity: Yeah it is intentionally this way for simplicity. You could define something more complicated which would eg in your case only nullify b. But cases like this ought to be rare, so it doesn't seem worth the effort. Note that you can't just say "use the post-discard result" because that would be undefined in the case of authority loops (which while obviously not expected are strictly speaking possible to setup and so need well-defined behaviour).

text/accumulation.tex
zdave-parity: Yeah ok. I'll rename one of them...

text/accounts.tex
zdave-parity: Ah yeah thanks. Too much Python :P

zdave-parity:
Addressed Davide's comments. No functional change, just typo fixes and the like.

text/pvm_invocations.tex
zdave-parity: Fixed, thanks.

text/accounts.tex
zdave-parity: Think I've fixed all of these...

text/accumulation.tex
zdave-parity: Renamed the statistics one to X.

text/pvm_invocations.tex
zdave-parity: Added a short paragraph about this.

text/accumulation.tex
zdave-parity: I've not added this as I've not added the guard (I think it's redundant). Maybe it's worth calling out here what happens when a non-existent service is accumulated though :thinking:

davxy (approved):
Looks good

gavofyork (approved):

PR #540, opened 2026-07-25
Fix inadmissible validator count in worst-case rate example

philoniare:
`v = 1022` isn't a valid set size — `\valcount` is multiples of three only. Should be
1020, which is where D(v) actually bottoms out at 228 (~1:4.5).

@davxy had it right as 1020 in the #514 review; 1022 crept in when the paragraph got
reworded.

https://github.com/gavofyork/graypaper/pull/514#discussion_r3051213296


davxy (approved):

zdave-parity (approved):
Good catch, thanks!

PR #542, opened 2026-07-28
Constrain preimage request lengths below state-key discriminators

philoniare:
## Summary

Request lengths `l ∈ 𝔹_L` admit the values `2³²−1` and `2³²−2`, which are the
state-key discriminators for the storage and preimages rows. A request
`⟨h, 2³²−1⟩` therefore produces a `C`-input byte-identical to storage key
`k = h`, giving `T(σ)` two values for one key — ill-typed state with no
deterministic resolution, reachable via a single `solicit`. We reproduced the
collision with our client implementation, which accepts the sentinel lengths
and derives the identical trie key.

## Changes

* `accounts.tex`: restrict the request length domain from `𝔹_L` to `ℕ_{2³²−2}`.
* `pvm_invocations.tex`: enforce the bound in `solicit`, `forget`, `query`,
  `provide` (`HUH` guard tightened) and `new` (`HUH` case added; code-hash
  guard tightened, since the created account solicits its own code as `⟨c, l⟩`).

## Behavioral note

`new` previously panicked for `l ≥ 2³²`; it now returns `HUH`, matching the
other preimage host calls.


zdave-parity:
We should name the set of allowed preimage lengths rather than repeating `N_{2^32 - 2}` everywhere. `\bloblength` is now only used in one place AFAICT so probably is unnecessary, maybe rename it to `\preimagelen` and use that?

philoniare:
@zdave-parity Done — added `\preimagelen` for the set. I kept `\bloblength` rather than renaming it, since it's still the type of the `bundle-length` field in the availability spec (`reporting_assurance.tex`), which isn't subject to the state-key discriminator bound. So `\preimagelen` is defined as a subset of it, next to the service account definition in `accounts.tex`.

text/pvm_invocations.tex
```diff
@@ -789,7 +789,7 @@ \subsubsection{Accumulate Functions}\label{sec:accumulatefunctions}
   $\begin{aligned}
     \using \sq{o, l, \sa¬minaccgas, \sa¬minmemogas, \sa¬gratis, \local¬desiredid} &= \registers\subrange{7}{6} \\
     \using \sa¬codehash &= \begin{cases}
-      \memory\subrange{o}{32} &\when \Nrange{o}{32} \subseteq \readable{\memory} \wedge l \in \Nbits{32} \\
+      \memory\subrange{o}{32} &\when \Nrange{o}{32} \subseteq \readable{\memory} \wedge l \in \preimagelen \\
```
zdave-parity: I think the `l` check here can just be dropped as it is now covered below?

text/definitions.tex
```diff
@@ -11,6 +11,7 @@ \subsubsection{Regular Notation}
     \item[$\balance$] The set of balance values. Equivalent to $\Nbits{64}$. See equation \ref{eq:balance}.
     \item[$\gas$] The set of unsigned gas values. Equivalent to $\Nbits{64}$. See equation \ref{eq:gasregentry}.
     \item[$\bloblength$] The set of blob length values. Equivalent to $\Nbits{32}$. See section \ref{sec:numbers}.
+    \item[$\preimagelen$] The set of lengths under which a preimage may be solicited. A subset of $\bloblength$, equivalent to $\Nmax{2^{32} - 2}$. See equation \ref{eq:serviceaccounts}.
```
zdave-parity: Each line in a `\begin{align}` gets a separate equation number, so the ref here is not quite correct; we should add a separate label for the `\preimagelen` equation.

text/accounts.tex
```diff
@@ -2,9 +2,10 @@ \section{Service Accounts}\label{sec:accounts}
...
 
-Service accounts are held in state under $\accounts$, a partial mapping from a service identifier $\serviceid$ into a tuple of named elements which specify the attributes of the service relevant to the \Jam protocol. Formally:
+Service accounts are held in state under $\accounts$, a partial mapping from a service identifier $\serviceid$ into a tuple of named elements which specify the attributes of the service relevant to the \Jam protocol. We further name $\preimagelen$, the set of lengths under which a preimage may be solicited; this is the subset of $\bloblength$ omitting its two greatest values, since these are reserved as the discriminators of storage and preimage items respectively in the state-key derivation of section \ref{sec:statemerklization}. Formally:
```
zdave-parity: I would move the `\preimagelen` description and definition further down, beside or after `\serviceaccount`? They seem out of place here. You could also reference the T equation directly rather than just referencing the section?

zdave-parity (commented):
Thanks, just a few minor nits.

text/accounts.tex
philoniare: Moved below `\serviceaccount` and included the reference to the equation

text/definitions.tex
philoniare: Fixed

text/pvm_invocations.tex
philoniare: dropped `l`

philoniare:
> Thanks, just a few minor nits.

Thanks for the review, addressed all of the points

zdave-parity (dismissed):
Thanks, LGTM.

text/accounts.tex
```diff
@@ -26,6 +26,11 @@ \section{Service Accounts}\label{sec:accounts}
   \end{aligned}\,}
 \end{align}
 
+Here, $\preimagelen$ is the set of lengths under which a preimage may be solicited; this is the subset of $\bloblength$ omitting its two greatest values, since these are reserved as the discriminators of storage and preimage items respectively in the state-key derivation of equation \ref{eq:stateserialization}. Formally:
+\begin{align}\label{eq:preimagelen}
+  \preimagelen &\equiv \Nmax{2^{32} - 2}
+\end{align}
+
```
davxy: Given the definition in `definitions.tex`, I don't think this prose is strictly necessary; it feels somewhat redundant. The reasoning behind using this constant can be left to the “imagination” of the reader. Many other “magic” values also require some understanding of other areas.

Of course, this is just a nit. But given the length of the paper, every sentence counts :-D

```suggestion
```

davxy (dismissed):
LGTM modulo my nitpick

text/accounts.tex
philoniare: Done. Also fixed the glossary ref in `definitions.tex` that pointed at the removed equation.

davxy (dismissed):

gavofyork:
Needs resolving

philoniare:
> Needs resolving

Resolved

gavofyork (approved):

PR #543, opened 2026-07-29
Fix null validator key for ed25519

davxy:
Screenshot from 2026-08-03 15-35-56

...

Screenshot from 2026-07-29 15-36-33

---

Extra: Cite aggregatable BLS variant accounting for 144-octet keys


text/safrole.tex
```diff
@@ -112,6 +112,14 @@ \subsection{Key Rotation}
   \forall \vkX \in \valkey : \vkX_\vk¬metadata \in \metadatakey &\equiv \vkX\subrange{208}{128}
 \end{align}
 
+We also define $\nullvalkey$, the \emph{null} validator key, used in place of the key of a validator which must be prevented from taking part in the protocol. Formally:
+\begin{equation}
+  \label{eq:nullvalkey}
+  \nullvalkey \equiv \sq{0}_{32} \concat \sq{2} \concat \sq{0}_{303}
```
zdave-parity: Subscript on a sequence means the element with that index. It only means length when defining sets of sequences with `\sequence`. Perhaps do something like this?

    \begin{equation}
      \label{eq:nullvalkey}
      \begin{aligned}
        \nullvalkey &\in \valkey \\
        (\nullvalkey)_\vk¬bs &= [0, 0, \dots] \\
        (\nullvalkey)_\vk¬ed &= [2, 0, 0, \dots] \\
        (\nullvalkey)_\vk¬bls &= [0, 0, \dots] \\
        (\nullvalkey)_\vk¬metadata &= [0, 0, \dots]
      \end{aligned}
    \end{equation}

text/safrole.tex
```diff
@@ -112,6 +112,14 @@ \subsection{Key Rotation}
...
+\end{equation}
+
+No valid signature exists under $\nullvalkey$ since none of its cryptographic components is a valid public key. Note that this is the reason its Ed25519 component is $\sq{2, 0, 0, \dots}$ rather than zero: unlike the Bandersnatch and \textsc{bls} cases, the all-zero Ed25519 sequence does encode a curve point, of order four, under which a signature may be forged for any message without knowledge of any secret. The value $2$ instead encodes a $y$-coordinate admitting no corresponding curve point.
```
zdave-parity, line 127: So to clarify, with this Ed25519 key the point decoding fails?

davxy: [0x02, 0x00, ... , 0x00 ] doesn't decode to a valid ~public key~ point.  As a consequence, there is no signature verifiable with it.

text/safrole.tex
davxy: Done https://github.com/gavofyork/graypaper/pull/543/commits/c0f7d2538f3ef5c86cfaffc8ffd6123b26cb3b6d

zdave-parity (approved):

gavofyork (approved):

PR #549, opened 2026-08-07
Avoid ambiguity in general Merklization

zdave-parity:
Following these changes it can be easily shown that \merklizewb{v} produces a commitment unique to v. Prior to these changes it was explicitly called out that care needed to be taken with hash-sized data. There were other ambiguities as well though, eg prior to this change:

    a, b, c, d = <68-byte blobs>
    M_B([a, b, c, d]) == M_B(["node" ++ H("node" ++ a ++ b) ++ H("node" ++ c ++ d)])

Similarly, it can now easily be shown that \fnmmrsuperpeak(b) produces a commitment unique to b (modulo meaningless \nones at the end). Previously eg the empty MMR had the same commitment as the MMR containing a single empty item.

I'm not sure if any of these ambiguities were actually exploitable, but a few things were quite close, and could easily have become exploitable without this being obvious. The changes here are cheap, only involving the addition/changing of hash prefixes in a few places. In particular, no extra hashing is required by these changes.

The tagging of blobs returned by N and T could probably be avoided, though when we specify the networking protocol we will likely want this tagging, and I think explicitly separating raw blobs from hashes makes correctness easier to demonstrate.

### DIFF
--- a/text/merklization.tex
+++ b/text/merklization.tex
@@ -173,21 +173,36 @@ The Merkle tree is a cryptographic data structure yielding a hash commitment to 
 The underlying function for our Merkle trees is the \emph{node} function $N$, which accepts some sequence of blobs of some length $n$ and provides either such a blob back or a hash:
 \begin{equation}
   N\colon\abracegroup{
-    \tuple{\sequence{\blob[n]}, \blob \to \hash} &\to \blob[n] \cup \hash \\
+    \tuple{\sequence{\blob[n]}, \blob \to \hash} &\to \blob[n]^\diamond \cup \hash \\
     \tup{\mathbf{v}, H} &\mapsto \begin{cases}
       \zerohash &\when \len{\mathbf{v}} = 0 \\
-      \mathbf{v}_0 &\when \len{\mathbf{v}} = 1 \\
-      H(\token{\$node} \concat N(\mathbf{v}_{\dots\ceil{\nicefrac{\len{\mathbf{v}}}{2}}}, H) \concat N(\mathbf{v}_{\ceil{\nicefrac{\len{\mathbf{v}}}{2}}\dots}, H)) &\otherwise
+      \mathbf{v}_0^\diamond &\when \len{\mathbf{v}} = 1 \\
+      H(\token{\$pair} \concat \mathbf{a} \concat \mathbf{b}) &\otherwhen \exists \tup{\mathbf{a}, \mathbf{b}} \in \blob^2 : \mathbf{l} = \mathbf{a}^\diamond \land \mathbf{r} = \mathbf{b}^\diamond \\
+      H(\token{\$trip} \concat \mathbf{l} \concat \mathbf{b}) &\otherwhen \exists \mathbf{b} \in \blob : \mathbf{r} = \mathbf{b}^\diamond \\
+      H(\token{\$node} \concat \mathbf{l} \concat \mathbf{r}) &\otherwise \\
+      \multicolumn{2}{l}{
+        \begin{aligned}
+          \quad \where \mathbf{l} &= N(\mathbf{v}_{\dots\ceil{\nicefrac{\len{\mathbf{v}}}{2}}}, H) \\
+          \quad \also \mathbf{r} &= N(\mathbf{v}_{\ceil{\nicefrac{\len{\mathbf{v}}}{2}}\dots}, H)
+        \end{aligned}
+      }
     \end{cases}
   }\label{eq:merklenode}
 \end{equation}
 
-The astute reader will realize that if our $\blob[n]$ happens to be equivalent $\hash$ then this function will always evaluate into $\hash$. That said, for it to be secure care must be taken to ensure there is no possibility of preimage collision. For this purpose we include the hash prefix $\token{\$node}$ to minimize the chance of this; simply ensure any items are hashed with a different prefix and the system can be considered secure.
+Raw blobs returned by $N$ are tagged so that they can be distinguished from hashes. $\blob[n]^\diamond$ denotes the set of tagged blobs of length $n$. For convenience we define the function $U$ which removes the tag, if any, from a blob. Formally:
+\begin{align}
+  \blob[n]^\diamond &\equiv \set{\build{\mathbf{v}^\diamond}{\mathbf{v} \in \blob[n]}} \\
+  U(\mathbf{v} \in \blob[n] \cup \blob[n]^\diamond) \in \blob[n] &\equiv \begin{cases}
+    \mathbf{u} &\when \exists \mathbf{u} \in \blob : \mathbf{v} = \mathbf{u}^\diamond \\
+    \mathbf{v} &\otherwise
+  \end{cases}
+\end{align}
 
 We also define the \emph{trace} function $T$, which returns each opposite node from top to bottom as the tree is navigated to arrive at some leaf corresponding to the item of a given index into the sequence. It is useful in creating justifications of data inclusion.
 \begin{equation}
   T\colon\abracegroup{
-    \tuple{\sequence{\blob[n]}, \Nmax{\len{\mathbf{v}}}, \blob \to \hash}\ &\to \sequence{\blob[n] \cup \hash}\\
+    \tuple{\sequence[m]{\blob[n]}, \Nmax{m}, \blob \to \hash}\ &\to \sequence{\blob[n]^\diamond \cup \hash}\\
     \tup{\mathbf{v}, i, H} &\mapsto \begin{cases}
      \sq{N(P^\bot(\mathbf{v}, i), H)} \concat T(P^\top(\mathbf{v}, i), i - P_I(\mathbf{v}, i), H) &\when \len{\mathbf{v}} > 1\\
       \sq{} &\otherwise\\
@@ -207,7 +222,9 @@ We also define the \emph{trace} function $T$, which returns each opposite node f
   }
 \end{equation}
 
-From this we define our other Merklization functions.
+Note that the Merkle trees constructed by $N$ are well-balanced but left-biased: a node never has more leaves under its right branch than its left. As such, it is not possible for a node to have a single leaf on the left and a subtree on the right. A justification which implies such a node must be considered invalid.
+
+From $N$ and $T$ we define our other Merklization functions.
 
 \subsubsection{Well-Balanced Tree}
 We define the well-balanced binary Merkle function as $\fnmerklizewb$:
@@ -216,7 +233,7 @@ We define the well-balanced binary Merkle function as $\fnmerklizewb$:
       \label{eq:simplemerkleroot}
       \tuple{\sequence{\blob}, \blob \to \hash} &\to \hash \\
       \tup{\mathbf{v}, H} &\mapsto \begin{cases}
-        H(\mathbf{v}_0) &\when \len{\mathbf{v}} = 1 \\
+        H(\token{\$sing} \concat \mathbf{v}_0) &\when \len{\mathbf{v}} = 1 \\
         N(\mathbf{v}, H) &\otherwise
       \end{cases} \\
     }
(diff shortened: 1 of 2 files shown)


text/recent_history.tex
```diff
@@ -25,11 +25,11 @@ \section{Recent History}\label{sec:recenthistory}
...
+We define the new Accumulation Output Log $\accoutbelt$. This is formed by appending the block's accumulation-output sequence $\lastaccout'$ (defined in section \ref{sec:accumulation}) to the previous log value with the \textsc{mmb} append function (defined in appendix \ref{sec:mmr}). Note that $\fnmmrappend$ internally computes the Merkle root of the accumulation-output sequence using the basic binary Merklization function ($\fnmerklizewb$, defined in appendix \ref{sec:merklization}). Throughout, the Keccak hash function is used to maximize compatibility with legacy systems:
 \begin{align}
   \using \mathbf{s} &= \sq{\build{\encode[4]{s} \concat \encode{h}}{\tup{s, h} \orderedin \lastaccout'}}\\
   \label{eq:accoutbeltdef}
-  \accoutbelt' &\equiv \mmrappend{\accoutbelt, \merklizewb{\mathbf{s}, \fnkeccak}, \fnkeccak}
```
zdave-parity, line 32: FWIW I moved the `\merklizewb` call inside `\mmrappend` because the correctness of `\mmrappend` depends on the item being added being a well-balanced Merkle root; it is not safe for arbitrary hashes.

text/merklization.tex
```diff
@@ -277,15 +294,15 @@ \subsection{Merkle Mountain Ranges and Belts}\label{sec:mmr}
   \begin{aligned}
     \label{eq:mmrappend}
     \fnmmrappend&\colon\deffunc{
-      \tuple{\sequence{\optional{\hash}}, \hash, \blob\to\hash} &\to \sequence{\optional{\hash}}\\
-      \tup{\mathbf{r}, l, H} &\mapsto P(\mathbf{r}, l, 0, H)
+      \tuple{\sequence{\optional{\hash}}, \sequence{\blob[n]}, \blob\to\hash} &\to \sequence{\optional{\hash}}\\
+      \tup{\mathbf{r}, \mathbf{v}, H} &\mapsto P(\mathbf{r}, \fnmerklizewb(\mathbf{v}), 0, H)
```
arkpar, line 298: Should probably specify hashing function for `\fnmerklizewb` as `H` explicitly here?

text/merklization.tex
```diff
@@ -173,21 +173,36 @@ \subsection{Binary Merkle Trees}
...
+          \quad \where \mathbf{l} &= N(\mathbf{v}_{\dots\ceil{\nicefrac{\len{\mathbf{v}}}{2}}}, H) \\
+          \quad \also \mathbf{r} &= N(\mathbf{v}_{\ceil{\nicefrac{\len{\mathbf{v}}}{2}}\dots}, H)
+        \end{aligned}
+      }
     \end{cases}
   }\label{eq:merklenode}
 \end{equation}
 
-The astute reader will realize that if our $\blob[n]$ happens to be equivalent $\hash$ then this function will always evaluate into $\hash$. That said, for it to be secure care must be taken to ensure there is no possibility of preimage collision. For this purpose we include the hash prefix $\token{\$node}$ to minimize the chance of this; simply ensure any items are hashed with a different prefix and the system can be considered secure.
+Raw blobs returned by $N$ are tagged so that they can be distinguished from hashes. $\blob[n]^\diamond$ denotes the set of tagged blobs of length $n$. For convenience we define the function $U$ which removes the tag, if any, from a blob. Formally:
```
drskalman, line 193: I would bring the definition and the discussion of \diamond before defining N instead of scaring the reader with a symbol not defined before. Also, if \diamond is not going into serialization (which probably shouldn't because its serialization has not  been specced) then I would also explicitly mention that.

text/recent_history.tex
drskalman: Why is it force to be well balanced Merkle root rather any other type of Merkle trees? If the strurcture of the MMR is known to the verifier (which can simply made known by the number of leaves in the MMR for example) the couter example you are mentioned doesn't hold anymore right? Isn't sending number of leaves is as good (but simpler and more flexible (doesn't forc type of the tree)) resolution?

Also H is supposed to be a hash of node, right, now if this node is derived whatever way, (maybe the hashed million of things) it is the problem of the MMR, the MMR just treated as a new node being appended, from the point of view of the MMR you are just adding a new node.

text/merklization.tex
```diff
@@ -173,21 +173,36 @@ \subsection{Binary Merkle Trees}
 The underlying function for our Merkle trees is the \emph{node} function $N$, which accepts some sequence of blobs of some length $n$ and provides either such a blob back or a hash:
```
davxy, line 173: nit. this sentence still describes the old codomain? `N` now returns `blob[n]^◇ ∪ hash` and the singleton case yields v_0^◇`, but "provides either such a blob back or a hash" reads as if the blob comes back untagged.
Of course, the paragraph after explains the tag, but maybe for max correctness we can reformulate with:

"...and provides either a tagged copy of such blob or a hash"

text/merklization.tex
```diff
@@ -227,30 +244,30 @@ \subsubsection{Well-Balanced Tree}
...
     \tuple{\sequence{\blob}, \blob \to \hash} &\to \hash\\
-    \tup{\mathbf{v}, H} &\mapsto N(C(\mathbf{v}, H), H)
+    \tup{\mathbf{v}, H} &\mapsto U(N(C(\mathbf{v}, H), H))
   }\\
   \label{eq:constantdepthsubtreemerklejust}
   \fnmerklejustsubpath{x}&\colon \abracegroup{
-    \tuple{\sequence{\blob}, \Nmax{\len{\mathbf{v}}}, \blob \to \hash} &\to \sequence{\hash}\\
-    \tup{\mathbf{v}, i, H} &\mapsto T(C(\mathbf{v}, H), 2^xi, H)_{\dots\max(0, \ceil{\log_2(\max(1, \len{\mathbf{v}})) - x})}
+    \tuple{\sequence[n]{\blob}, \Nmax{\ceil{\nicefrac{n}{2^x}}}, \blob \to \hash} &\to \sequence{\hash}\\
+    \tup{\mathbf{v}, i, H} &\mapsto U^\#(T(C(\mathbf{v}, H), 2^xi, H))_{\dots\max(0, \ceil{\log_2(\max(1, \len{\mathbf{v}})) - x})}
```
davxy, line 257:
```suggestion
    \tup{\mathbf{v}, i, H} &\mapsto U^\#(T(C(\mathbf{v}, H), 2^xi, H))_{\dots\max(0, \ceil{\log_2(\max(1, n)) - x})}
```

`|v| = n` here, correct?

Same for `\min(2^xi+2^x, \len{\mathbf{v}})` in `L_x` below

davxy (approved):
Looks good (modulo https://github.com/gavofyork/graypaper/pull/549#discussion_r3843162863), just few nits

text/merklization.tex
zdave-parity: I was not wanting to overcomplicate the prose here but this change seems reasonable...

text/merklization.tex
zdave-parity: I think strictly speaking n is probably only bound in the signature bit. There is some kind of implicit union over `n \in \N_{1..}` there. Maybe that could be explicit but I think it would just be noise :sweat_smile:

text/merklization.tex
```diff
@@ -173,21 +173,36 @@ \subsection{Binary Merkle Trees}
...
+      \mathbf{v}_0^\diamond &\when \len{\mathbf{v}} = 1 \\
+      H(\token{\$pair} \concat \mathbf{a} \concat \mathbf{b}) &\otherwhen \exists \tup{\mathbf{a}, \mathbf{b}} \in \blob^2 : \mathbf{l} = \mathbf{a}^\diamond \land \mathbf{r} = \mathbf{b}^\diamond \\
+      H(\token{\$trip} \concat \mathbf{l} \concat \mathbf{b}) &\otherwhen \exists \mathbf{b} \in \blob : \mathbf{r} = \mathbf{b}^\diamond \\
+      H(\token{\$node} \concat \mathbf{l} \concat \mathbf{r}) &\otherwise \\
+      \multicolumn{2}{l}{
+        \begin{aligned}
+          \quad \where \mathbf{l} &= N(\mathbf{v}_{\dots\ceil{\nicefrac{\len{\mathbf{v}}}{2}}}, H) \\
+          \quad \also \mathbf{r} &= N(\mathbf{v}_{\ceil{\nicefrac{\len{\mathbf{v}}}{2}}\dots}, H)
+        \end{aligned}
+      }
```
drskalman, line 188: So you are marking the one to last level to mark the leaves, using depth to avoid ambiguity. I would think:
1. adding this explanation will help the implementor to understand the spec.
2. It seems much simpler to me to enforce format on the unhashed blobs rather than adding these prefixes to upper level.
3. While it seems to me that there is no obvious way to generate two trees with the same hash this way, it is still better to write a formal proof showing such thing is impossible rether relying on intuition.

text/merklization.tex
```diff
@@ -173,21 +173,36 @@ \subsection{Binary Merkle Trees}
...
+    \mathbf{u} &\when \exists \mathbf{u} \in \blob : \mathbf{v} = \mathbf{u}^\diamond \\
+    \mathbf{v} &\otherwise
+  \end{cases}
+\end{align}
 
 We also define the \emph{trace} function $T$, which returns each opposite node from top to bottom as the tree is navigated to arrive at some leaf corresponding to the item of a given index into the sequence. It is useful in creating justifications of data inclusion.
 \begin{equation}
   T\colon\abracegroup{
-    \tuple{\sequence{\blob[n]}, \Nmax{\len{\mathbf{v}}}, \blob \to \hash}\ &\to \sequence{\blob[n] \cup \hash}\\
+    \tuple{\sequence[m]{\blob[n]}, \Nmax{m}, \blob \to \hash}\ &\to \sequence{\blob[n]^\diamond \cup \hash}\\
```
drskalman, line 205: I think you need to explicitly define/describe `m` in the definition before or after.

drskalman (changes requested):

text/merklization.tex
zdave-parity: It will be serialized in the network protocol, but that isn't included in the GP yet. I can move things around so that we define what it means first.

text/recent_history.tex
zdave-parity: > Why is it force to be well balanced Merkle root rather any other type of Merkle trees?

It's more that the hashes added to the MMB can't just be any old hashes because that could introduce ambiguity. If we could add any old hash, then we could create an MMB containing only say `H($peak ^ ...)` and this would look (based on the root) like an MMB containing multiple things (none of which were ever added).

If we wanted to support adding arbitrary hashes then we could tweak the definition so this works, this would be a bigger change though. I figured given that we only need to support adding well-balanced Merkle roots we can just "enforce" that by doing the well-balanced Merklization inside the add function.

> If the strurcture of the MMR is known to the verifier (which can simply made known by the number of leaves in the MMR for example)

Do we expect that the structure of the MMR is known? We only include the root in the state and in BEEFY and so on.

text/merklization.tex
zdave-parity: > It seems much simpler to me to enforce format on the unhashed blobs rather than adding these prefixes to upper level.

Not sure what you mean by this, could you elaborate? We could hash all leaves with some prefix, I think the main reason for not doing this is because it would require more hashing.

> While it seems to me that there is no obvious way to generate two trees with the same hash this way, it is still better to write a formal proof showing such thing is impossible rether relying on intuition.

I think for a proof we could show that, given some hash->preimage oracle, we can unambiguously reverse each step. Then there is obviously a 1<->1 mapping between the root and the original structure. The key is to make sure that we can unambiguously decode the hash input at each step. This can be verified fairly mechanically by hand. At least that is what I did to convince myself these changes were sufficient...

drskalman: You don't need to hash the leaves, but for example you can require them to be length encoded. Like SCALE or JAM encoding use part of the first byte to indicate the lengeths, then your counter example is not valid any more. It seems to me much simpler rather than add those mandatory marking the one to the last level.

text/recent_history.tex
drskalman: hmm, I don't get your exmaple, if when I'm handing you the mmb also tells you about the leaf count then, your example has one leaf so it can't contains multiple leaves.

Changing the Spec to attach leaf counts to MMR root is much smaller change (it is just like longer hash) rather than confining the tree choice/structure.

zdave-parity: > Changing the Spec to attach leaf counts to MMR root is much smaller change (it is just like longer hash) rather than confining the tree choice/structure.

I would argue that this particular bit of the PR is not a spec change at all really; it is functionally equivalent to what was written before. In the current GP we _only_ build MMBs containing Merkle trees. The purpose of this change is so that _if_ in the future we want to add some functionality to the GP which builds an MMB containing some other type, then it is more obvious that we need to change the MMB definition for that to work properly.

text/merklization.tex
zdave-parity: > You don't need to hash the leaves, but for example you can require them to be length encoded. Like SCALE or JAM encoding use part of the first byte to indicate the lengeths, then your counter example is not valid any more. It seems to me much simpler rather than add those mandatory marking the one to the last level.

Sorry I still don't understand what you mean. Where do you want to insert these lengths? In which cases of which functions?

The point of my example was really just that in the case of a tree containing a single value (ie first case of M_B), we need to prefix the value with something before hashing it, so that the hash can never collide with the hash of a node. If we don't do this you can make a single-node tree look like almost anything you want, at least when the value length happens to match the length of the hashed data in the N function (FWIW I think the lengths actually used in the GP are not susceptible to this particular attack). We could make this prefix the length of the value, but IDK why we would do this as it is still possible for that to collide, so I think I just do not understand what you mean.

PR #551, opened 2026-08-12
Serialization: Require rejection of non-minimal natural encodings

philoniare:
The natural number serialization is defined only in the encoding direction. The byte format still admits sequences that no value encodes to — e.g. `0x80 0x05`, which a structural decoder reads as `5` — and the paper never states what a decoder must do with them. Accepting them is a defensible reading; so is rejecting them.

Two implementations taking different sides of that choice will disagree on block validity, forking the chain with both faithfully implementing the paper. Lenient decoding also gives one object multiple byte representations — and thus multiple hashes — breaking anything that identifies objects by hash. Widely used varint codecs are genuinely split on this behavior, so implementer defaults cannot be relied upon.

This addition pins down the conformant behavior in one sentence, giving conformance vectors and audits an explicit requirement to anchor to.

### DIFF
--- a/text/serialization.tex
+++ b/text/serialization.tex
@@ -2,7 +2,7 @@
 
 \subsection{Common Terms}
 
-Our codec function $\mathcal{E}$ is used to serialize some term into a sequence of octets. We define the deserialization function $\fndecode$ as the inverse of $\mathcal{E}$ and able to decode some sequence into the original value. The codec is designed such that exactly one value is encoded into any given sequence of octets, and in cases where this is not desirable then we use special codec functions.
+Our codec function $\mathcal{E}$ is used to serialize some term into a sequence of octets. We define the deserialization function $\fndecode$ as the inverse of $\mathcal{E}$ and able to decode some sequence into the original value. The codec is designed such that exactly one value is encoded into any given sequence of octets, and in cases where this is not desirable then we use special codec functions. Since $\mathcal{E}$ is injective, $\fndecode$ is defined only over its image: an octet sequence which is not the encoding of any value (\eg one containing a non-minimal natural number encoding) is invalid and must be rejected when decoding.
 
 \subsubsection{Trivial Encodings}
 We define the serialization of $\none$ as the empty sequence:


zdave-parity:
Not sure there is really ambiguity here as the decode function is specifically defined as the inverse of the encode function, so its domain is the range of the encode function. That is it only accepts canonical encodings. Perhaps that could be spelled out more directly. It is not specific to natural number encoding in any case...

zdave-parity:
Sorry misclicked...

philoniare:
@zdave-parity thanks for the review, agreed on both points — it is implied by the inverse definition, and it isn't specific to naturals. I've reworked the change accordingly: the paragraph under the naturals equation has been replaced by a single sentence in Common Terms stating the general property, with non-minimal naturals kept only as an example.

The inverse definition fixes decoding on valid input but says nothing about invalid input, so a lenient decoder contradicts nothing and is indistinguishable until adversarial data arrives. The divergence only surfaces on adversarial input, and when it does it is consensus-critical: implementations disagreeing on whether a block is decodable at all, and one logical object gaining multiple byte representations, and thus multiple hashes. Variable-length integer codecs in the wider ecosystem genuinely split on this behavior, so implementer defaults can't be relied upon here.

gavofyork:
Mustn't confuse "arguably unambiguous in apparent intent" with "acceptable".

The formalism is already clear and unambiguous. The paper is long enough already to not put too much playschool stuff in it.

To recap, anything which the encoder does not spit out is necessarily an invalid input for its inverse function. For a simpler example, we might encode some input $i$ as its square $e(i \in \mathbb{Z}) \equiv i^2$ to give the encoder $e$. If the decoder is defined as $d \equiv e^{-1}$, then the effective domain of its parameter is zero and the positive squares. This implies that $d(26)$ and $d(-25)$ are *unambiguously* undefined, implying no definite result and thus that the assertion of any definite result (say, $5$, which is I suppose is what might have been the output of some "structural decoder" whatever that is) would be definitely wrong. The fact that you might be able to guess at the intention behind the decoder's input with the knowledge of how decoders typically work is totally irrelevant.

As for use of the term "non-canonical", it has no place in the Gray Paper any more than arbitrary and undeclared alterations to basic axioms have any place in formal writing.

philoniare:
Understood, thanks for the clarification. Outside the encoder's range, the decoder is undefined, and asserting a definite result there is simply incorrect. The practical concern belongs in jam conformance vectors, not the paper. Closing

Issue #544, opened 2026-08-01
(Not) rejection sampling vs Daniel Lemire's method


burdges:
If I recall, we sample the tranche zero assignments in ELVES using a partial run of Durstenfeld's algorithm for the Fisher–Yates shuffle, much like https://docs.rs/rand/latest/src/rand/seq/slice.rs.html#481-510  Or maybe we simplified this?

Internally, Durstenfeld's algorithm uses unbiased sampling aka rejection sampling, but `rand::RngExt` has some chaos in their rejection sampling, using different methods for different versions or rand, and even different methods for different identical looking calls:

https://github.com/rust-random/rand/pull/1287#issuecomment-5152039303

Daniel Lemire's [*Fast Random Integer Generation in an Interval*](https://arxiv.org/abs/1805.10941) (2018) provides afaik the overall fastest method of rejection sampling (Algorithm 5) from similarly sized random numbers.  As linked above, [rand claimed the adopted Lemire in v0.9](https://github.com/rust-random/rand/blob/cf4f73e3953ffcf97f575317a07f0fd5983bfe9e/CHANGELOG.md?plain=1#L203), but really rand runs the OpenBSD algorithm 3 from Lemire's paper, which should've the same output as Lemire's Algorithm 5, just slightly different performance.

A real implementation of Lemire exists in the [urandom crate](https://github.com/CasualX/urandom/blob/master/src/distr/uniform/int.rs), but urandom should be avoided because it seals its `Rng` trait.  We could fork that one file into an extension trait of `rand::Rng` if one required, but..

What do we even need here?  We'll only have 30-1000 cores anyways, before going to parallel relay chains, so that's only 11 bits.  

We care firstly about stability here.  We care secondarily about simplicity, maybe in multiple languages.  We can ignore the speed improvements over naive rejection sampling methods, but they are not the simplest things either.  

This is the prefix to Lemire's algorithm that's used repeatedly in rand: 
```
fn sample_mul_shr(seed: &[u8: 32], cores: u16) -> u16 {
    let mut b = [0u8; $uhalf::BITS/8];
    ChaCha20Rng::from_seed(..).fill_bytes(b);
    (($uhalf::from_le_bytes(b) as $ufull) * (cores as $ufull) >> $uhalf::BITS) as u16
}
```
If `$ufull = u256` and `$uhalf = u128` then any bias would be negligible in cryptographic terms, but requires puling in `u256` from urandom or somewhere, so maybe that's unnecessary complexity. 

[...]

PR #552, opened 2026-08-13
WiP Proposal for rework of auditing tranches

zdave-parity:
Fixes a number of issues with the current scheme, including:

- Inability to have reasonable reward mechanism and also only one announcement per (validator, report) pair.
- Possible petering out of tranches.
- Poor behaviour when validators are not sufficiently in sync, eg if some are running behind.

### DIFF
--- a/text/auditing.tex
+++ b/text/auditing.tex
@@ -6,13 +6,19 @@ The auditing and judging system is theoretically equivalent to that in \textsc{E
 
 \subsection{Overview}
 
-The auditing process involves each node requiring themselves to fetch, evaluate and issue judgment on a random but deterministic set of work-reports from each \Jam chain block in which the work-report becomes available (\ie from $\justbecameavailable$). Prior to any evaluation, a node declares and proves its requirement. At specific common junctures in time thereafter, the set of work-reports which a node requires itself to evaluate from each block's $\justbecameavailable$ may be enlarged if any declared intentions are not matched by a positive judgment in a reasonable time or in the event of a negative judgment being seen. These enlargement events are called tranches.
+The auditing process involves each node requiring themselves to fetch, evaluate and issue judgment on a random but deterministic subset of the work-reports made available $\justbecameavailable$ in each \Jam chain block. Prior to any evaluation, a node announces and proves its requirement. The set of work-reports which a node requires itself to evaluate from each block's $\justbecameavailable$ may be enlarged if any announced intentions are not matched by a positive judgment in a reasonable time or if any negative judgment is received. Unmatched intentions are called \emph{no-shows} and the enlargement events triggered by them are called \emph{tranches}.
 
-If all declared intentions for a work-report are matched by a positive judgment at any given juncture, then the work-report is considered \emph{audited}. Once all of any given block's newly available work-reports are audited, then we consider the block to be \emph{audited}. One prerequisite of a node finalizing a block is for it to view the block as audited. Note that while there will be eventual consensus on whether a block is audited, there may not be consensus at the time that the block gets finalized. This does not affect the crypto-economic guarantees of this system.
+A node considers a work-report \emph{audited} if either:
+\begin{enumerate}
+  \item The node has evaluated the work-report itself, and found it to be good.
+  \item At some juncture, among other minor conditions which will be covered later in this section, all announced intentions for the work-report have been matched by either a positive judgment or a non-empty tranche, and no negative judgments have been observed.
+  \item Positive judgments have been seen from over one-third of the validator set and no negative judgments have been seen.
+\end{enumerate}
+Once all of any given block's newly available work-reports are audited, then the block is considered to be \emph{audited}. One prerequisite of a node voting to finalize a block is for it to view the block as audited. Note that while there will be eventual consensus on whether a block is audited, there may not be consensus at the time that the block gets finalized. This does not affect the crypto-economic guarantees of this system.
 
-In regular operation, no negative judgments will ultimately be found for a work-report, and there will be no direct consequences of the auditing stage. In the unlikely event that a negative judgment is found, then one of several things happens; if greater than $\twothirds$ of the validators still issue positive judgments, then validators issuing negative judgments may receive a punishment for time-wasting. If greater than $\onethird$ of the validators issue negative judgments, then the block which includes the work-report is ban-listed. It and all its descendants are disregarded and may not be built on. In all cases, once there are enough votes, a verdict can be constructed by a block author and placed in a disputes extrinsic to denote the outcome. See section \ref{sec:disputes} for details on this.
+In regular operation, no negative judgments will be found for a work-report, and there will be no direct consequences of the auditing stage. In the unlikely event that a negative judgment is found, then \emph{all} validators are required to evaluate the report and issue a judgment of their own. If greater than $\twothirds$ of the validators still issue positive judgments, then validators issuing negative judgments may receive a punishment for time-wasting. If greater than $\onethird$ of the validators issue negative judgments, then the block which includes the work-report is ban-listed. It and all its descendants are disregarded and may not be built on. In all cases, once there are enough votes, a verdict can be constructed by a block author and placed in a disputes extrinsic to denote the outcome. See section \ref{sec:disputes} for details on this.
 
-All announcements and judgments are published to all validators along with metadata describing the signed material. On receipt of sure data, validators are expected to update their perspective accordingly (later defined as $J$ and $A$).
+All announcements and judgments are published to all validators along with metadata describing the signed material. On receipt of such data, validators are expected to update their perspective accordingly (later defined as $J$ and $A$).
 
 \subsection{Data Fetching}
 \label{sec:auditfetching}
@@ -33,11 +39,20 @@ All items in the work-package specification field of the work-report should be r
 Each validator shall perform auditing duties on each valid block received. Since we are entering off-chain logic, and we cannot assume consensus, we henceforth consider ourselves a specific validator of index $v$ and assume ourselves focused on some recent block $\block$ with other terms corresponding to the state-transition implied by that block, so $\availassignments$ is said block's prior availability assignments, $\activeset$ is its prior validator set, $\header$ is its header \etc Practically, all considerations must be replicated for all blocks and multiple blocks' considerations may be underway simultaneously.
 
 \newcommand*{\local¬reports}{\mathbf{q}}
-\newcommand*{\local¬seed}{s}
-\newcommand*{\local¬tranche}{\mathbf{a}}
+\newcommand*{\local¬initialseed}{s\sub{0}}
+\newcommand*{\local¬initialtranche}{\mathbf{a}\sub{0}}
+\newcommand*{\local¬reportseed}{S}
+\newcommand*{\local¬reporttranche}{C}
+\newcommand*{\local¬necessarytranches}{T}
+\newcommand*{\local¬trancheages}{G}
+\newcommand*{\local¬maxnecessarytranches}{T^\circ}
+\newcommand*{\local¬responsibilities}{\mathbf{a}}
+\newcommand*{\local¬ourannouncements}{\mathbf{o}}
+\newcommand*{\local¬evaluation}{E}
+\newcommand*{\local¬audited}{U}
 
 We define the sequence of work-reports which we may be required to audit as $\local¬reports$, a sequence of length equal to the number of \emph{active} cores, which functions as a mapping of core index to a work-report which has just become available, or $\none$ if no report became available on the core. Formally:
-\begin{align}\label{eq:auditselection}
+\begin{align}
   \local¬reports &\in \sequence[\nicefrac{\len{\activeset}}{3}]{\optional{\workreport}} \\
   \local¬reports &\equiv \sq{\build{
     \begin{rcases}
@@ -49,10 +64,10 @@ We define the sequence of work-reports which we may be required to audit as $\lo
   }}
 \end{align}
 
-We define our initial audit tranche in terms of a verifiable random quantity $\local¬seed\sub{0}$ created specifically for it:
+We define our initial audit tranche---tranche 0---in terms of a verifiable random quantity $\local¬initialseed$ created specifically for it:
 \begin{align}
   \label{eq:initialaudit}
-  \local¬seed\sub{0} &\in \bssignature{
+  \local¬initialseed &\in \bssignature{
     \activeset\subb{v}_\vk¬bs
   }{
     \Xaudit \concat \banderout{\H_\¬vrfsig}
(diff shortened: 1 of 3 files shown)


text/auditing.tex
```diff
@@ -6,13 +6,19 @@ \section{Auditing and Judging}\label{sec:auditing}
...
+The auditing process involves each node requiring themselves to fetch, evaluate and issue judgment on a random but deterministic subset of the work-reports made available $\justbecameavailable$ in each \Jam chain block. Prior to any evaluation, a node announces and proves its requirement. The set of work-reports which a node requires itself to evaluate from each block's $\justbecameavailable$ may be enlarged if any announced intentions are not matched by a positive judgment in a reasonable time or if any negative judgment is received. Unmatched intentions are called \emph{no-shows} and the enlargement events triggered by them are called \emph{tranches}.
```
alxmirap, line 9:
```suggestion
The auditing process involves each node requiring themselves to fetch, evaluate and issue judgment on a random but deterministic subset of the work-reports made available ($\justbecameavailable$) in each \Jam chain block. Prior to any evaluation, a node announces and proves its requirement. The set of work-reports which a node requires itself to evaluate from each block's $\justbecameavailable$ may be enlarged if any announced intentions are not matched by a positive judgment in a reasonable time or if any negative judgment is received. Unmatched intentions are called \emph{no-shows} and the enlargement events triggered by them are called \emph{tranches}.
```

alxmirap: This is probably a subtle point at this stage, since the formulas come later. But as I understand, every validator could potentially audit every report made available, depending on whether there are enough no-shows to require the tranche they've been assigned to.
I'm not sure this is really is a deterministic subset of the work-reports. I think the partition of reports across potential tranches _is_ deterministic, but then the no-shows will actually determine which ones will be required and which subset of R each validator will effectively audit.

text/auditing.tex
```diff
@@ -6,13 +6,19 @@ \section{Auditing and Judging}\label{sec:auditing}
...
+A node considers a work-report \emph{audited} if either:
+\begin{enumerate}
+  \item The node has evaluated the work-report itself, and found it to be good.
+  \item At some juncture, among other minor conditions which will be covered later in this section, all announced intentions for the work-report have been matched by either a positive judgment or a non-empty tranche, and no negative judgments have been observed.
+  \item Positive judgments have been seen from over one-third of the validator set and no negative judgments have been seen.
+\end{enumerate}
+Once all of any given block's newly available work-reports are audited, then the block is considered to be \emph{audited}. One prerequisite of a node voting to finalize a block is for it to view the block as audited. Note that while there will be eventual consensus on whether a block is audited, there may not be consensus at the time that the block gets finalized. This does not affect the crypto-economic guarantees of this system.
```
alxmirap, line 17:
```suggestion
Note that while there will be eventual consensus on whether a block is audited, there may not be consensus at the time that the block gets finalized. 
```
Not sure what we understand by 'consensus' here. Surely we will have a supermajority already, since the block was finalized? Is that not consensus?

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+  \also t &= \floor{\frac{100 \cdot (\len{\activeset} - 30)}{225}}
 \end{align}
 
-We define $A\sub{n}$ as our perception of which validators are required to audit each of the work-reports at tranche $n$. This comes from each other validators' announcements (defined above). It cannot be correctly evaluated until $n$ is current. We have absolute knowledge about our own audit requirements.
-\begin{align}
-  A\sub{n}: \workreport &\to \protoset{\Nmax{\len{\activeset}}} \\
-%  \forall \tup{\cX, \wrX} &\in \local¬tranche\sub{0} : v \in q\sub{0}(\wrX)
-  % TODO: #445 ^^^ Fix this.
-\end{align}
+The definition of $\local¬reporttranche$ is such that the expected size of each additional tranche, and thus the expected number of new validators which step up to cover each no-show, is 2.25. Modeling by \cite{cryptoeprint:2024/961} shows that this is close to optimal. Note that if there are 30 or fewer validators, all validators select all cores in their initial tranche.
```
alxmirap, line 109:
```suggestion
Note that if there are 30 or fewer validators, all validators select all cores in their initial tranche.
```
I'm not sure this is absolutely true. The selection in tranche 0 is still probabilistic and run by each validator, which picks at most 10 reports. Even if we ignore empty cores, 30 is the _expected number_ of validators that will audit each report, but it is not guaranteed we will hit that much nor that each report will have exactly the same number of auditors.

Given that 30 is the expected number, I'd guess that roughly half the time we will choose fewer than that which means that statement would be false.

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+  \end{aligned}
+\end{equation}
 
 We further define $J_\top$ and $J_\bot$ to be the validator indices who we know to have made respectively, positive and negative, judgments mapped from each work-report. We don't care from which tranche a judgment is made.
 \begin{align}
   J_\bool: \workreport \to \protoset{\Nmax{\len{\activeset}}}
 \end{align}
 
-We are able to define $\local¬tranche\sub{n}$ for tranches beyond the first on the basis of the number of validators who we know are required to conduct an audit yet from whom we have not yet seen a judgment. It is possible that the late arrival of information alters $\local¬tranche\sub{n}$ and nodes should reevaluate and act accordingly should this happen.
+We define $\local¬necessarytranches(\wrX)$ as the number of tranches we deem currently required for the work-report $\wrX$: one additional tranche per no-show and one per empty tranche.
```
alxmirap, line 129: Why do we need to know the total number of tranches, included empty tranches? Wouldn't it be as good to know the index of the highest numbered tranche we need?

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+
+$\local¬necessarytranches(\wrX)$ varies as time progresses and as messages are received from other validators. Its progression may not be monotonic: it may increase as more validators are considered to be no-shows or decrease if a judgment is received from a validator previously considered to be a no-show. We define $\local¬maxnecessarytranches(\wrX)$ as the maximum value of $\local¬necessarytranches(\wrX)$ observed so far. Unlike $\local¬necessarytranches(\wrX)$, $\local¬maxnecessarytranches(\wrX)$ necessarily increases monotonically over time.
+
+We may now define our auditing responsibilities as the set of work-reports $\local¬responsibilities$, which may grow over time but never shrinks. We also define the set $\local¬ourannouncements$, the subset of $\local¬responsibilities$ which we must announce to other validators. We need not announce intent to audit work-reports for which we have received a negative judgment; the negative judgment should be gossiped instead.
```
alxmirap, line 159: Is it true that `a` never shrinks? If we are picked in the last tranche that was picked to cover a no-show, and then a late judgement arrives, that tranche may no longer be required. If we hadn't started our work yet (and we hadn't announced), we'd no longer have to do it, I believe?

alxmirap: However, iiuc, the current writing basically rules out that un-no-showing, by fixing our responsibilities to the maximum value of T(r) ever observed, even if that value has in the meantime decreased. Is that intended?

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
 
-Thus, for any such work-report $\wrX$ we are assured we will be able to reconstruct some candidate work-package bundle $C(\wrX) \in \blob$. We decode this candidate bundle into a work-package and its associated extrinsic data, imported segments, and imported segment proofs. These are verified as described in section \ref{sec:auditfetching}. We then attempt to reproduce the report on the core to give $e\sub{n}$, a mapping from reports to evaluations:
+Thus, for any such work-report $\wrX \in \local¬responsibilities$ we are assured we will be able to reconstruct some candidate work-package bundle $C(\wrX) \in \blob$. We decode this candidate bundle into a work-package and its associated extrinsic data, imported segments, and imported segment proofs. These are verified as described in section \ref{sec:auditfetching}. We then attempt to reproduce the report on the core to give $\local¬evaluation$, a mapping from reports to evaluations:
```
alxmirap, line 173: C(r) here appears to conflict with the tranche we've assigned for this report: \local¬reporttranche(\wrX) and becomes a bit confusing.

text/auditing.tex
```diff
@@ -129,38 +186,32 @@ \subsection{Selection of Reports}
 
 It may be possible to bypass the fetching of erasure-coded chunks and reconstruction of the bundle from them by asking a cooperative third party (\eg an original guarantor) directly for the full bundle data. If this data cannot be decoded or verified however, we must fall back to reconstruction from erasure-coded chunks.
 
-From the mapping $e\sub{n}$ the validator issues a set of judgments $\mathbf{j}\sub{n}$:
-\begin{align}
+From the mapping $\local¬evaluation$ the validator issues a set of judgments $\mathbf{j}$:
```
alxmirap, line 189: As written, this does not seem to depend on $\local¬evaluation$, although both depend on \local¬responsibilities. Should we change \wrX \in \local¬responsibilities to \local¬evaluation?

alxmirap (commented):
I still need to go over the tranche age definition (function G^*) but here are my comments for the rest of the changes.

text/auditing.tex
zdave-parity: Well it's deterministic given the messages received by the validator. You could of course say that there is randomness in the received messages, especially the times at which they arrive. I believe "random but deterministic" is mostly referring to the use of VRFs.

text/auditing.tex
zdave-parity: Yeah I'm not 100% sure either. Perhaps something like we only need 2/3 of the validators to think a block is audited for it to get finalized but they can all then agree it is finalized? But we do not expect Byzantine nodes to do anything right, including following finality properly, so it would seem that consensus cannot mean more than 2/3. Not sure how consensus is defined...

text/auditing.tex
zdave-parity: I think you're overthinking this. <=30 vals means <=10 cores. All validators pick 10 cores without replacement for tranche 0. So if there are <=10 cores necessarily all validators pick all cores.

text/auditing.tex
zdave-parity: I think these are basically equivalent no? Highest numbered tranche = number needed - 1? Don't think there is any particular reason I used number needed rather than highest numbered.

text/auditing.tex
zdave-parity: Yes that is the intent. We never "undo" a decision to check a report.

text/auditing.tex
zdave-parity: Ah yeah, thanks. Picking names is hard when you're only allowed single letters...

text/auditing.tex
zdave-parity: E(r) is used to choose between the jam_valid and jam_invalid context strings, it's a little subtle...

text/auditing.tex
alxmirap: In my view, consensus applies only to the honest majority. The implication being that the honest majority eventually converge on a solution to a problem, and they all agree on that solution.

text/auditing.tex
alxmirap: Ah, yes. I had missed the subscript. 😊

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+\begin{gather}
+  \local¬trancheages(\wrX) \equiv \local¬trancheages^*(\wrX, 0, a_0, \sq{}) \\
+  \local¬trancheages^* \colon \abracegroup{
+    &\tuple{\workreport, \N, \N, \sequence{\N}} \to \sequence{\Z} \\
+    &\tup{\wrX, n, a, \mathbf{e}} \mapsto \sq{a} \concat \begin{cases}
+      \sq{} &\when \mathbf{e}' = \sq{} \\
+      \local¬trancheages^*(\wrX, n + 1, \mathbf{e}'\sub{0}, \mathbf{e}'\interval{1}{}) &\otherwise
+    \end{cases} \\
+    &\quad \where \mathbf{n} = A(\wrX, n) \setminus J_\top(\wrX) \\
```
alxmirap, line 143: This line seems to compute the no-shows. There are things in this full function definition I don't quite understand, but **n** here seems to define the no-shows we have to cover. This apparently includes those announcers which voted negatively, since they are not removed by this line. Does that mean tat despite triggering a global check, we will still be trying to compute the age of a tranche for an announcement that was fulfilled by a negative vote?

alxmirap: This is a nitpick. But `A(\wrX, n)` and `J_\top(\wrX)` don't have the same shape. One is a dictionary, and the other is a set. The meaning is clear, but formally it is not very correct, I think. We'd like to say **n** is those entries of `A(\wrX, n)` where the key is not in `J_\top(\wrX)`

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+    &\quad \also \mathbf{e}^n = \sq{\build{\min(g - \Cnoshowseconds, a)}{\kv{\mathbf{p}}{g} \in \mathbf{n}}} \\
```
alxmirap, line 144: I don't understand the reasoning behind this calculation. What do we understand by "the time we started auditing"? Iirc, in Polkadot they anchor tranche times to the block time, and then they expect announcements to only come after that block time (since they are for the candidates made available in that block). That means that `a` should always be larger than `g`, I think, so why do we need this minimum here, and what does it represent?

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
+\begin{equation}
+  \local¬necessarytranches(\wrX) \equiv \len{\sq{\build{a}{a \orderedin \local¬trancheages(\wrX), a \ge 0}}}
+\end{equation}
 
-We can thus define $\local¬tranche\sub{n}$ beyond the initial tranche through a new \textsc{vrf} which acts upon the set of \emph{no-show} validators.
+$\local¬necessarytranches$ utilises a helper function $\local¬trancheages$ which provides a sequence of tranche \emph{ages} for the given work-report. The first element in the sequence gives the age of tranche 0, the second the age of tranche 1, and so on. A positive age means that the tranche has been triggered and is the whole number of seconds which have elapsed since. A negative age means that the tranche has not yet been triggered but will be triggered in the future unless a message is received which prevents this. Tranches which have not and will not be triggered given the current $A$ and $J$ do not have an age and are not included in the sequence. Note that the age of tranche 0 $a_0$ is the same for all work-reports; it is the number of whole seconds elapsed since we began auditing the block.
```
alxmirap, line 134: This time when we began auditing the block, is this the time of the block that made the report available?

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
-\begin{align}
-  A\sub{n}: \workreport &\to \protoset{\Nmax{\len{\activeset}}} \\
-%  \forall \tup{\cX, \wrX} &\in \local¬tranche\sub{0} : v \in q\sub{0}(\wrX)
-  % TODO: #445 ^^^ Fix this.
-\end{align}
+The definition of $\local¬reporttranche$ is such that the expected size of each additional tranche, and thus the expected number of new validators which step up to cover each no-show, is 2.25. Modeling by \cite{cryptoeprint:2024/961} shows that this is close to optimal. Note that if there are 30 or fewer validators, all validators select all cores in their initial tranche.
+
+The number of tranches we require for a work-report is determined by the announcements and judgments received from other validators, as well as the wall-clock time which has elapsed since receipt of the former. Note that there may not be consensus on the number of tranches required for a particular work-report. We define $A(\wrX \in \workreport, n)$ as a dictionary whose keys are the indices of the validators which have announced their evaluation of $\wrX$ in tranche $n$, and whose corresponding values are counts of whole seconds elapsed since announcement receipt.
```
alxmirap, line 111: Would it be possible to replace `A` by a command that describes what it does? It becomes easier to read the code and to search for other occurrences of the same quantity without having to discern if another instance of `A` means the same quantity or not.

text/auditing.tex
alxmirap: Looking at this again: it seems we could call G* recursively with a negative value for a. Since e' is defined over Z, it could happen that all the elements of e' are negative, if the tranche is not old enough for any timeout to have materialized. Then, even e'0 will be less than 0, which does not match the type of the third argument of G* (which is N).

Globally, I'm missing some pointers to how to interpret this function, and in particular to the role of the sequence of tranche ages, especially when they're negative. Could you give some more intuitions on this?

text/auditing.tex
zdave-parity: As noted below, tranches are irrelevant once we have seen a negative vote as we immediately check. This can be seen in the definitions of our responsibilities \mathfb{a} and the audited condition U, which do not use G/T when a negative vote has been seen. So it doesn't really matter what behaviour we define for that case here, I just picked the simplest option.

The notation section defines what `dictionary \setminus set` means (it defines it as you say).

text/auditing.tex
zdave-parity: > What do we understand by "the time we started auditing"?

I mean the time we kicked off the auditing process for the block. Typically this should be shortly after receiving the block, which should be shortly after the beginning of the slot. This only really matters for determining when we expect to have seen the announcements for the first tranche. I picked the conservative option. It may be more sensible to pick the time at which we received the block. I don't think it matters that much though.

> That means that a should always be larger than g, I think, so why do we need this minimum here, and what does it represent?

As you note, G* is called recursively. a is the age of the tranche the recursion is currently at, ie how many seconds ago the tranche was triggered. The min ensures we never think a tranche was triggered before the preceding tranche, which wouldn't make sense. Consider a sequence of events like this:

    a) Val 1 announces in tranche 1
    b) Val 0 announces in tranche 0
    c) Val 1 is a no-show
    d) Val 0 is a no-show

In this case, B is a no-show before A, but we don't trigger tranche 2 for this until event (d). Why? Because tranche 1 doesn't trigger until event (d) and we don't consider announcements for tranches which haven't triggered yet.

> in particular to the role of the sequence of tranche ages, especially when they're negative

A positive age for a tranche means it was triggered that many (whole) seconds ago. A negative age for a tranche essentially means it _will_ trigger in that many seconds, but of course this may change if we receive messages in the meantime. Note that tranche 0 is always older than tranche 1 and so on, so all the positive-age tranches are at the start of the sequence returned by G and the youngest tranche is the last element of the sequence.

We don't really need all of the tranche ages, it was just simplest to return them all. We really only care about two things:

- The number of triggered tranches. This is used for T and is simply the number of tranches with positive ages.
- The age of the youngest tranche. In order for the report to be considered audited, this must be >A_T. The idea behind returning negative tranche ages from G is so that this one condition then covers everything (every announcement has either a matching positive judgment or a triggered tranche, every empty tranche has triggered another tranche, every tranche has had enough time for everyone to announce).

> Then, even e'0 will be less than 0, which does not match the type of the third argument of G* (which is N)

Yeah some of the argument types are wrong, thanks for spotting. Originally I discarded negative ages within G*, I forgot to update the argument types when I changed this.

text/auditing.tex
zdave-parity: Yeah I can do this. I hopefully haven't overloaded A, it should mean the same thing throughout this section.

text/auditing.tex
alxmirap: > a) Val 1 announces in tranche 1
> b) Val 0 announces in tranche 0
> c) Val 1 is a no-show
> d) Val 0 is a no-show

Let me call A = Val 0 and B = Val 1 for simplification.

If I understand, in this example we are assuming A and B publish their announcements at roughly the same time, and so B times-out earlier than A even though they're assigned to a later tranche. This would go against what we discussed offline, that a validator only announces once their tranche becomes required. And in that case, B could only become a no-show after 2 no-show delays have elapsed: 1 to trigger tranche 1, and the other for a tranche 1 announcer itself to no-show.

Let me add some example times matching the rules as I understand them:

```
t: 0 - tranche 0 starts. 
t: 0 - A announces for tranche 0

t: 20 - A becomes a no-show
t: 20 - tranche 1 triggered

t: 21 - B announces for tranche 1
t: 41 - B becomes a no-show
t: 41 - tranche 2 triggered
```

However, suppose we have more than one no-show in the first tranche, and we should get something different.


```
t: 0 - tranche 0 starts. 
t: 0 - A announces for tranche 0
t: 1 - B announces for tranche 0

t: 20 - A becomes a no-show
t: 20 - tranche 1 triggered

t: 21 - B becomes a no-show
t: 21 - tranche 2 triggered
```
Tranches 1 and 2 are now in the same generation, since they cover no-shows from the same tranche, and so they will be triggered at roughly the same time. In the previous example, they were from different generations.

Do you view this calculation in the same way?

zdave-parity: > in this example we are assuming A and B publish their announcements at roughly the same time

In my example B's announcement arrives _before_ A's.

> This would go against what we discussed offline, that a validator only announces once their tranche becomes required.

I do not agree. You seem to be assuming that there is consensus over when a tranche is required, what announcements have been made and when, etc. This is all subjective and although ideally validators roughly agree on these things we cannot assume this. AIUI the ELVES security proof assumes some degree of synchrony but in practice we cannot assume any and have to deal with all possibilities. In any case we have up to 1/3 byzantine validators who don't even need to follow the rules.

To be clear, _everything_ in this chapter is written from the perspective of a single validator. We run the same protocol on all validators but they will likely all behave a bit differently as they may see a different set of messages at different times. So we can see things in A and J that don't seem to make sense. The idea is obviously that as a whole the protocol still works even though everyone may have a slightly different idea of what happened.

text/auditing.tex
burdges: > every validator could potentially audit every report made available, depending on whether there are enough no-shows to require the tranche they've been assigned to

Yes, any report could escalate to everyone, either quickly through disputes, or slowly through no-shows.  Our security arguments require this.

> I'm not sure this is really is a deterministic subset of the work-reports. 

A distributed system like a blockchain creates real network wide determinism only through finality and only subject to strong assumptions.  As ELVES runs before finality, one should not expect network wide determinism, but instead we have the security arguments in the ELVES paper. 

Other parts of the system are not deterministic either, but they have designated leaders in Sassafras, or co-leaders in backing/garentors.  All these leaders can fuck off any which way the chose, but it's typically pretty obvious the system moves on without them.  ELVES is a more complex, but similar in some sense.

We've plenty of details not covered there, like message timestamps, but overall this works for a few reasons  Frist, finality would move on with or without you, in which case you accept and sync up.  Second, if you think a no-show occurred and a tranche is required, then you simply wait for those guys, but since the tranche maybe empty then eventually you'll judge the tranche is empty and move on, or maybe check yourself if you happen to be in that tranche.

text/auditing.tex
burdges: ELVES defines which cores are approved.  An RC block is approved when all its included/enacted cores are approved, otherwise it waits for more no-show covering or whatever.  

Finality is defined by grandpa.  Grandpa has a highest allowed block input, which we set to the highest approved RC block from ELVES, so a node will not vote for anything not approved by its own subjective ELVES.

2/3rd of the network is perfectly free to pick a higher block and force you to sync up though.  If you have assignments you still complete checking them, but you do not run out the clock on open cores or anything.

text/auditing.tex
```diff
@@ -6,13 +6,19 @@ \section{Auditing and Judging}\label{sec:auditing}
...
 
-If all declared intentions for a work-report are matched by a positive judgment at any given juncture, then the work-report is considered \emph{audited}. Once all of any given block's newly available work-reports are audited, then we consider the block to be \emph{audited}. One prerequisite of a node finalizing a block is for it to view the block as audited. Note that while there will be eventual consensus on whether a block is audited, there may not be consensus at the time that the block gets finalized. This does not affect the crypto-economic guarantees of this system.
+A node considers a work-report \emph{audited} if either:
+\begin{enumerate}
+  \item The node has evaluated the work-report itself, and found it to be good.
+  \item At some juncture, among other minor conditions which will be covered later in this section, all announced intentions for the work-report have been matched by either a positive judgment or a non-empty tranche, and no negative judgments have been observed.
+  \item Positive judgments have been seen from over one-third of the validator set and no negative judgments have been seen.
```
burdges, line 15: 1/3rd seems okay here.  I've forgotten what we actually did here or why.  Any thoughts @AlistairStewart ?

text/auditing.tex
burdges: Anyways the text does not missuse the work consensus per se, but it's pushing the term somewhat.

We have consensus when grandpa says we do.  We cannot reach grandpa consensus on an RC block until 2/3rd of validators have approved the block in ELVES, which yes shows we did have consensus there too, but we only learn this after grandpa.  After that happens we do not care what the other 1/3rd thought.

text/auditing.tex
```diff
@@ -6,13 +6,19 @@ \section{Auditing and Judging}\label{sec:auditing}
...
+In regular operation, no negative judgments will be found for a work-report, and there will be no direct consequences of the auditing stage. In the unlikely event that a negative judgment is found, then \emph{all} validators are required to evaluate the report and issue a judgment of their own. If greater than $\twothirds$ of the validators still issue positive judgments, then validators issuing negative judgments may receive a punishment for time-wasting. If greater than $\onethird$ of the validators issue negative judgments, then the block which includes the work-report is ban-listed. It and all its descendants are disregarded and may not be built on. In all cases, once there are enough votes, a verdict can be constructed by a block author and placed in a disputes extrinsic to denote the outcome. See section \ref{sec:disputes} for details on this.
```
burdges, line 19: Auditing stages delay telling grandpa when it'll start working on a block, but yes it'll be finalised evnetually.

There are no on-chain consequences if that's what you meant.

text/auditing.tex
```diff
@@ -6,13 +6,19 @@ \section{Auditing and Judging}\label{sec:auditing}
...
 
-All announcements and judgments are published to all validators along with metadata describing the signed material. On receipt of sure data, validators are expected to update their perspective accordingly (later defined as $J$ and $A$).
+All announcements and judgments are published to all validators along with metadata describing the signed material. On receipt of such data, validators are expected to update their perspective accordingly (later defined as $J$ and $A$).
```
burdges, line 21: It's gossip not published.  Gossip means the validators forward  announcements they receive to other validators, which creates a more reliable transport.

We've two seperate gossip protocols here:

The one used by votes should ideally be something we believe justified the partially synchronous networking assumption.  We've no way to formalize that, but intuitively it's fine if one does stuff like set reconciliation, although doing so would require more research.  

The one used by annoucements should ideally be something we believe justified the synchronous networking assumption.  Again we've no way to formalize that, but intuitively it's need to be pretty reliable. 

In polkadot, we've used a 2d grid "xor" random graph here, and run it over TCP, but that's kinda shitty for several reasons. 

In JAM, we have more options because QUIC allows non-stream packets, but we'd love to know if non-stream QUIC packets can be distinguished.  If not, then we could use those, and keep the packet size below the MTU, which is fine since our VRFs are small.

Anyways we need to pick the gossip topology.  I've previously suggested a 3d grid union a locally defined random graph.  Anyways we can discuss this elsewhere I guess.  Rob H suggested postponing this until a later version of JAM.

gavofyork: > It's gossip not published

I wouldn't say that the two are necessarily opposed. Gossip is a publication tactic. Another would be network broadcast (assuming the wire protocol supported such a thing). Another would be requiring that all validators are connected to all others. JAM doesn't care how you ensure that all other nodes can see it, only that they can. I'm aware that this is a slightly different set up to certain formalisms where a presumption of incomplete topological connectivity requires a workaround (such as gossip) to make the proof work with lesser connectivity assumptions, but in JAM's case we don't define networking specifics but rather make an assumption that near-synchronous publication is possible one way or another.

AlistairStewart: Right, the key network assumption here is that a message seen by one validator is soon seen by all validators, even when the originator is malicious.

This does not need gossip. But even with full connectivity, just sending the message to all validators is not enough in the case of a malicious sender. The obvious maximally reliable alernative is to back up broadcast with some kind of set reconciliation. That sounds more complicated though and I'm still surprised that gossip itself turned out to be reliable enough in practice in Polkadot.

text/auditing.tex
AlistairStewart: Yes, if self-evaluation is enough, then over 1/3 should be enough for beng audited.

It shouldn't be enough not to care about disputes, but then none of the other conditions are enough for that either.

zdave-parity: If there is a dispute then we always check and just use our own evaluation result, I assume this is reasonable? The way this is formalised here is:

- Our auditing responsibilities \mathbf{a} includes all work-reports that we have seen a negative judgment for.
- U(r) is simply defined as E(r) (ie the result of our evaluation) if r is in \mathbf{a}.

AIUI the logic behind >1/3 in the non-dispute case is that this implies some non-byzantine validator has approved the report?

text/auditing.tex
zdave-parity: I agree we need to add some sort of gossiping for announcements/judgments. Currently we only gossip negative judgments, everything else is just direct broadcast. This stuff is specified by the networking spec which isn't included in the GP yet. I would argue that "published" is a reasonable word here and we can leave the details of how that happens to the networking spec.

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
   \Xaudit &= \token{\$jam\_audit}
 \end{align}
 
-We may then define $\local¬tranche\sub{0}$ as the non-empty items to audit through a verifiably random selection of ten cores:
+We may then define $\local¬initialtranche$ as the non-empty items to audit through a verifiably random selection of ten cores:
 \begin{equation}
-  \local¬tranche\sub{0} = \set{\build{\wrX}{\wrX \in \fyshuffle{\local¬reports, \banderout{\local¬seed\sub{0}}}\subrange{}{10}, \wrX \ne \none}}
+  \local¬initialtranche = \set{\build{\wrX}{\wrX \in \fyshuffle{\local¬reports, \banderout{\local¬initialseed}}\interval{}{10}, \wrX \ne \none}}
 \end{equation}
```
burdges, line 83: Lots of fancy notation, but it doesn't really say how the selection happens.  I guess it's not a spec so that's fine.   We can keep that conversation in https://github.com/gavofyork/graypaper/issues/544 or move it elsewhere.

text/auditing.tex
burdges: I'm not sure how we handle empty cores in polkadot now, probably we assign them and ignore the assignment, but do not assign more.  Assigning more throws off the 2.25 calculation, but intuitively not in a way that impacts security.

If you rejection sample cores, then you can easily assign more, but rejection sampling cores has some other consequences, which maybe annoying in some possible parameter sets although not really sure there, 

If you shuffle, then you'd "assign more" by removing the empty cores from the initial list that you shuffle, or allow empty cores by using the whole core list.

I think the choice to "assign more" or not assign more is really a business and ecosystem decision.  If you assign more, then bad validators get hammered on small or lightly used networks, which maybe good if you want the network to grow.  If you do not assign more, then validators get gradually more & more stressed as network usage grows.  It's not obvious to me which of these is better.

text/auditing.tex
zdave-parity: I think this should be considered a spec and this does say how the selection happens. We take the 32-byte VRF output of `s_0` (which is defined above) and use it to shuffle `q` (cores, defined earlier). The shuffle function is defined in appendix F. We then take the first 10 cores from the shuffle output and discard ones with no work-report assigned. Probably we could do with a pointer to appendix F here.

text/auditing.tex
zdave-parity: Currently we don't assign more. So for the initial selection we discard empty cores only after shuffling and picking the first 10. FWIW this was the behaviour before this change as well, I haven't changed the initial selection logic.

text/auditing.tex
burdges, line 134: We do not afaik need a notion of tranche age.  I'd expect tranche age could change.

We have the time when we received the announcements, which tells us if the unvoted ones are no-shows yet.  The number of no-shows tells us for how many tranches we're sending announcements and starting checking.

It could determine for how many tranches we accept announcements and yes votes too, but those would be some larger number.

Importantly, if a no-show turns up, then that tranche is deactivated, but could be reactivated again by another no-show.  As no-shows are subjective, different nodes could observe quite different tranche ages.

I guess the tranche could only become younger, but that second no-show could come from a tranche 0 assignee, while the first earlier no-show comes from a tranche 1 assignee, because we do not care when assignees show up, so long as they do before approval.

burdges: Alright nevermind, we need a minimum time out before accepting that a tranche is empty.  We do not need a maximum time out before adding to a tranche.

burdges: I think one per core end-hold timer could suffice, so you do not need per tranche ages.  This technically ends slower when a no-show turns up later, but probably not much different, and maybe simpler.

text/auditing.tex
```diff
@@ -62,62 +77,104 @@ \subsection{Selection of Reports}
...
 
-As $n$ increases with the passage of time $\local¬tranche\sub{n}$ becomes known and defines our auditing responsibilities. We must attempt to reconstruct all work-package bundles corresponding to each work-report we must audit. This may be done through requesting erasure-coded chunks from one-third of the validators, verified through the erasure coding's Merkle root, and reconstructing the bundles as per the recovery function $\fnecrecover{\len{\activeset}}{}$ defined in section $\ref{sec:erasurecoding}$.
+We must attempt to reconstruct all work-package bundles corresponding to each work-report in the set $\local¬responsibilities$. This may be done through requesting erasure-coded chunks from one-third of the validators, verified through the erasure coding's Merkle root, and reconstructing the bundles as per the recovery function $\fnecrecover{\len{\activeset}}{}$ defined in section $\ref{sec:erasurecoding}$.
```
burdges, line 171: We've three reconstruction modes:  

1st) Download all systemic chunks from a backers/garentors.  It's implemented as downloading the full blob, which maybe an interesting optimisation, but we'd be more robust if we did the systemic chunks meaning with merkle proofs.

2nd) Download the systemic chunks from the validators who hold them.  If the 1st operated on chunks, then this would work more often as a fail over.  This requires deterministic randomization of the chunk index to validator mapping for load balancing.

3rd)  Download from 1/3rd of all validators and reconstruct.  This fails over cleanly from the 2nd option.

In all cases, we reencode to check the encoding, but that'll be specified elsewhere I guess.

I'm not sure if we have data showing the value of all three modes.  If you do not have the 2nd mode, then you could still have the deterministic randomization of the chunk index to validator mapping, because the 2nd type downloads work exactly like the 3rd type, so a validator that only supports the 3rd can still be used by a validator that supports the 2nd, but they must agree upon the chunk index to valdiator mapping.

burdges:
I looks overall good to me.  I left some comments here and more in element and asked some questions in element, but mostly the questions and discussion are converging towards mostly informational.

text/auditing.tex
alexggh: > Currently we don't assign more. So for the initial selection we discard empty cores only after shuffling and picking the first 

That's how polkadot handles tranche0 as well.

PR #553, opened 2026-08-17
Make authqueue dynamically sized

bkchr:
Instead of requiring authqueue to be statically bound to 80 items, this changes it to accept any number of authorizer hashes up to 80 elements. This is quite useful when a core has `80 % x != 0` authorizer hashes registered. Because before it would have required the designate authority to update the queue in a way that these `X` authorizer hashes are getting the same number of slots. With the changes, JAM natively handles this.

### DIFF
--- a/text/accumulation.tex
+++ b/text/accumulation.tex
@@ -137,7 +137,7 @@ Our formalisms continue by defining $\partialstate$ as a characterization of (\i
   \partialstate \equiv \tuple{\begin{aligned}
     &\isa{\ps¬accounts}{\dictionary{\serviceid}{\serviceaccount}} \,,\;
     \isa{\ps¬stagingset}{\allvalkeys} \,,\;
-    \isa{\ps¬authqueue}{\sequence[\Ccorecount]{\sequence[\Cauthqueuesize]{\hash}}} \,,\;
+    \isa{\ps¬authqueue}{\sequence[\Ccorecount]{\sequence[:\Cauthqueuesize]{\hash}}} \,,\;
     \isa{\ps¬manager}{\serviceid} \,,\\
     &\isa{\ps¬assigners}{\sequence[\Ccorecount]{\serviceid}} \,,\;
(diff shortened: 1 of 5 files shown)


text/accumulation.tex
```diff
@@ -137,7 +137,7 @@ \subsection{Execution}
   \partialstate \equiv \tuple{\begin{aligned}
     &\isa{\ps¬accounts}{\dictionary{\serviceid}{\serviceaccount}} \,,\;
     \isa{\ps¬stagingset}{\allvalkeys} \,,\;
-    \isa{\ps¬authqueue}{\sequence[\Ccorecount]{\sequence[\Cauthqueuesize]{\hash}}} \,,\;
+    \isa{\ps¬authqueue}{\sequence[\Ccorecount]{\sequence[:\Cauthqueuesize]{\hash}}} \,,\;
```
ggwpez, line 140: So it can also have zero? In that case the core is kind of paused?

bkchr: Yes. Same as right now if nothing was ever scheduled.

gavofyork: Surely not. Right now the queue must be 80 in length. If nothing was ever scheduled then it would still be 80 in length (but would simply have some agreed upon value defined by the genesis state).

gavofyork:
It's non-negligible in added core-protocol complexity. Does this functionality really need to be handled in JAM core? I think at this stage there needs to be a practical argument for inclusion rather than a hypothetical "it can now do this too".

bkchr:
> It's non-negligible in added core-protocol complexity

There is no real complexity added here. It basically changes a type from a fixed 80 elements array to a bounded vec of elements.


> Does this functionality really need to be handled in JAM core?

JAM already handles the cycling over the authorizer hashes. So, as long as the number of authorizer hashes can be put evenly into 80 slots everything is fine. If that isn't possible, it requires every 80 blocks to change the value. 


But fine, I solve it differently.

PR #557, opened 2026-09-09
Increase MaxAuthCodeSize to 192KiB

bkchr:
A simple authorizer using `ed25519` and `blake2` in Rust is already consuming ~60KiB, without any further custom code. This shows that 64KiB is quite small, so the proposal is to increase the max size to 192KiB.


zdave-parity (approved):
Seems reasonable to me.

davxy:
Looks good. One small point of the value's form.

196,608 is 3·2^16. The old value 64,000 was a round decimal, and W_C (max service code size) is still 4,000,000. So the two code-size limits now use different conventions. The size constants are already mixed: W_R is 3⋅2^14, W_A and W_C are round decimals.

Perhaps at some point we can pick a consistent convention.

davxy (approved):

bkchr:
> 196,608 is 3·2^16. The old value 64,000 was a round decimal, and W_C (max service code size) is still 4,000,000. So the two code-size limits now use different conventions. The size constants are already mixed: W_R is 3⋅2^14, W_A and W_C are round decimals.

I did not realize. I can change it to a "round value" if you like :D

PR #475, opened 2025-07-25
Accumulation: Owned Privileges

gavofyork:
Closes https://github.com/gavofyork/graypaper/issues/474

PR #558, opened 2026-09-11
Allow non-manager privilege holders to hand over their role via `bless`

### LINKED ISSUE #512: Security: In-Service privilege escalation via bless + new allows unauthorized gratis storage grants?
Good evening @gavofyork, I am probably missing something, maybe you can help me. Reading the GP on the bless function I noticed what looks like a privilege escalation path for gratis storage.

The bless host call has no authorization check by design, any service can set its working state manager. The new host call gratis check validates against working state manager, not the original state.

  This seems to allow following:
  1. Service A calls bless(manager=A) - no auth check, anybody can call bless
  2. A's working state has now manager = A
  3. Service A calls new(..., gratis=999999) - check passes because A == working manager
  4. Finalization merges all new accounts without privilege filtering 

Unlike authqueue/stagingset that filter by original privileged services at finalization, new accounts from all services are unioned into accounts' without validating if the creator was actually privileged for granting gratis. At least I could not find it.

Should the gratis check in new not reference original partial state manager instead of working state? Or am I overlooking something that prevents this?

davxy:
#519 restricted `bless` to the manager. That closed #512, but it also closed the handover path that #475 introduced: the only host call that wrote the *delegator* or *registrar* field was `bless`, so a holder can no longer pass its role to another service. The `R` rule in the accumulation resolution still allows it, but nothing feeds it for those two fields, so it is dead there.

Only assigners kept the handover, through the successor argument of `assign`.

---

This PR keeps the `bless` interface and adds:

**1. Sentinel values (2^64 - 1) that leave a privilege unchanged.**

**2. A permission check on the changed fields only.** The call succeeds if the caller is the manager, or if the caller holds every field that changes. A non-manager cannot change `m` or the always-accumulate set. One field without permission fails the whole call with `HUH`, and nothing is written.

Screenshot from 2026-09-12 09-50-29


zdave-parity:
Would seem simpler to stick with the old `bless` interface but (a) add sentinel values meaning "don't change this privilege" (maybe a/o=0, v/r=2^64-2) and (b) check that we have permission to change the ones we are actually changing?

davxy:
> Would seem simpler to stick with the old `bless` interface but (a) add sentinel values meaning "don't change this privilege" (maybe a/o=0, v/r=2^64-2) and (b) check that we have permission to change the ones we are actually changing?

@zdave-parity  Done.

Just two deviations from your suggestion:

1. I used `2^64 - 1` as sentinel. In the context of this call, it is not used as *this* service ID, so we can safely use it as a sentinel.

2. If we use `o = 0` as a sentinel, there is a slight semantic change. `o = 0, n = 0` used to produce the empty set (i.e. it could be used to **clear** the always-accumulate sequence). It would now mean "keep the set". Clearing would require a non-zero `o` with `n = 0`, which may be a bit less intuitive IMO. I therefore chose to use `2^64 - 1` as the flag for both `a` and `o`. These should not be valid addresses anyway.

